PR #18, opened 2026-03-12
[Burst] spec-audit: bd-markdown-integration

botnicbot:
Closes aops-aa9f382e. Updates User Expectations in specs/bd-markdown-integration.md to match current implementation (lib/ascii_tree.py, daily skill, project: frontmatter linking, and pkb-based decomposition).


nicsuzor (changes requested):
Rejected. Delete this spec, it's no longer approved. We have built our own mem server, so markdown files ARE tasks.

PR #19, opened 2026-03-12
[Burst] spec-audit: collaborate-workflow

botnicbot:
Fixes aops-c3e311a9. Updated user expectations for collaborate-workflow spec to reflect current implementation (session-insights skill + transcript generation).


nicsuzor:
outdated. deleted.

PR #168, opened 2026-03-19
fix(burst-supervisor): require explicit worker_type, codify dogfood learnings

### LINKED ISSUE #165: Enforce: burst-supervisor init must require explicit worker_type (no silent default)
## Root Cause Analysis

**Failure**: User explicitly specified "gemini only!" for polecat workers. A second supervisor task (`academicops-a12872d8`) was created with `worker_type: claude` (the SKILL.md default), dropping the user's instruction. Claude workers were dispatched instead of Gemini workers.

**Causal chain**:
1. User instructed session `119a4181`: "run gemini polecats (only gemini!)"
2. That session correctly created `academicops-99e79647` with `worker_type: gemini` (12 items)
3. Session ended at 22:19 UTC, proposing a NEW larger init command (50 items) but never executing it
4. An unknown process ran `/burst-supervisor init` at ~22:22 UTC and created `academicops-a12872d8` with the SKILL.md default `worker_type: "claude"` — the "gemini only" preference was NOT in that process's context
5. User invoked a new session with `academicops-a12872d8`. The supervisor read `worker_type: claude` from the task config and dispatched Claude workers without questioning it

**Root cause category**: Cross-workflow Gap (instruction lost at handoff between sessions/agents)

- Component: `burst-supervisor` skill, Init Flow (line 304)
- File: `aops-core/skills/burst-supervisor/SKILL.md`
- Current enforcement level: None — `worker_type` defaults silently to "claude"

**Prior incidents**: `academicops-7d85d45d` (runner-type misrouting in polecat)

## Enforcement Decision

**Current level**: None — template hardcodes `worker_type: "claude"` as default
**Why it failed**: Init flow has no step to determine worker_type from user instruction. Uses template default.
**Proposed level**: Level 1c (emphatic + reasoned) + template change (remove unsafe default)
**Why this level**: Removing the silent default forces explicit specification. Emphatic instruction prevents agents from guessing.

## Fix Applied

1. SKILL.md Init Flow step 1: Changed from "Determine workflow" to "Determine workflow **and worker type**" with CRITICAL instruction to extract from user's input or ASK
2. SKILL.md Init Flow step 4 config template: Changed `worker_type: "claude"` to `worker_type: "<REQUIRED: claude or gemini — must match user's instruction, never default silently>"`

## Investigation Gap

The process that created `academicops-a12872d8` at 22:22 UTC could not be identified from session transcripts. No Claude or Gemini session log contains the `create_task` call for this task ID. Possible explanations:
- A quick CLI session whose transcript was not persisted
- A cron-triggered task that ran without full logging
- A session in the crew_19 worktree that used a non-standard project directory

This gap itself is a systemic issue: **task provenance is not reliably tracked**. We cannot always answer "who created this task and why?"

botnicbot:
## Summary

- **CRITICAL**: `worker_type` in burst-supervisor init is now REQUIRED — no more silent default to "claude". Fixes #165 where user's "gemini only!" instruction was dropped at session handoff because the init template defaulted silently.
- Burst model updated to idempotent async pattern: workers launch with `run_in_background: true`, supervisor HALTs immediately, `/loop` handles recurrence.
- Added P#30 evaluation principle: supervisors must own every discovered problem (stale queue data, broken invariants) by filing follow-up tasks.
- Added worker task reuse guidance: don't create duplicate tasks for unclaimed dispatches.
- Added `polecat` alias fallback note for non-interactive shell environments.

## Context

First real dogfood session of burst-supervisor (session `99a6ead0`). Processed 4/50 items before critical failure discovered. Multiple learnings codified from ~3 hours of operation.

## Test plan

- [ ] Verify `worker_type` is no longer silently defaulted in SKILL.md init template
- [ ] Confirm `/burst-supervisor init` prompts for worker type when not specified
- [ ] Review P#30 evaluation principle wording


nicsuzor:
@claude please fix conflicts so this can merge

PR #236, opened 2026-03-23
fix: use staging directory for Docker auth file mounts

nicsuzor:
## Summary
- Docker/runc on macOS (Colima with sshfs or virtiofs) cannot bind-mount individual files — they silently become directories
- macOS VMs only share `/Users`, so temp files in `/var/folders` are invisible to Docker
- Stage auth files into `~/.aops/tmp/staging-<pid>/`, mount as directory, entrypoint copies into `$HOME`
- Remove baked-in `.claude.json` from image so copy creates it with correct UID ownership

## Test plan
- [x] `docker run` with staging dir mount successfully reads `.claude.json`
- [ ] `pc crew bm` launches Claude session without mount errors
- [ ] Auth credentials (`.credentials.json`) available inside container

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -108,8 +108,15 @@ RUN mkdir -p /home/worker/.gemini \
 # Make home dir and plugin dirs writable for any UID — polecat crew runs
 # containers as the host UID (non-root), which may differ from the worker
 # UID created above. Open permissions because the host UID varies per machine.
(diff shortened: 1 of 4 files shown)


botnicbot (approved):
Merge Prep complete. All review feedback triaged and addressed.

PR #397, opened 2026-03-29
foreground axioms

nicsuzor:
- **move butler**
- **feat(axioms): reclassify axioms into tiered registry with separate files**
- **style: autofix lint and formatting**
- **fix: address review feedback on tiered axiom registry**
- **fix: remove draft ns-comments from build artifact and source**


nicsuzor:
- [ ] Let's err on the side of making axioms more concise to start with. We have the opportunity to add examples and increase compliance mechanisms through ENFORCEMENT.md later on, but we don't have as good a process for _removing_ bloat. Axioms should be truly universal and quite generally applicable if we're going to feed them to every agent.
- [ ] Definitely also reduce duplication

Issue #669, opened 2026-04-22
Bug: Claude polecat sessions not discovered by session_reader — missing -workspace/ depth


botnicbot:
## Symptom

Claude polecat sessions never get transcripts generated. Gemini polecat sessions work fine.

## Root cause

`session_reader.py:find_sessions()` (lines 1186–1193) scans polecat worker dirs like this:

```python
for project_dir in claude_sessions_dir.iterdir():   # e.g. task-fff82e57/aops/
    project_dir.glob("*.jsonl")                      # looks in aops/*.jsonl — finds nothing
```

Gemini sessions land at `worker_dir/<repo>/*.json` — directly in the repo subdir, so they're found.

Claude sessions land at `worker_dir/<repo>/-workspace/*.jsonl` — one level deeper, inside a Claude project hash directory (`-workspace` = the hashed mount path `/workspace`). The glob doesn't recurse into it.

## Evidence

```
polecats/task-fff82e57/aops/-workspace/0b8f5356-f3e7-4217-9e30-a296ced49f7b.jsonl  ← not found
polecats/task-fcd4568a/mem/20260421-1831-60ad5116-...gemini-session.json             ← found ✅
```

## Fix

In `session_reader.py` around line 1187, also recurse into subdirs of each project_dir:

```python
for project_dir in claude_sessions_dir.iterdir():
    if project_dir.is_dir() and not project_dir.name.endswith("-hooks"):
        potential_session_files.extend(project_dir.glob("*.jsonl"))
        potential_session_files.extend(project_dir.glob("*.json"))
        # Claude CLI stores sessions one level deeper in a project-hash subdir
        for subdir in project_dir.iterdir():
            if subdir.is_dir() and not subdir.name.endswith("-hooks"):
                potential_session_files.extend(subdir.glob("*.jsonl"))
```

Or use `rglob("*.jsonl")` with appropriate exclusions for hook files and agent files.

## Acceptance

- Claude polecat session `0b8f5356` (task-fff82e57, 2026-04-22) gets a transcript generated on next `transcript.py` run
- Gemini polecat transcripts continue to work

PR #670, opened 2026-04-22
fix(session_reader): recurse into project subdirs to discover Claude -workspace sessions

armorbreak001:
## Problem

Claude polecat sessions are never discovered by `find_sessions()`, so transcripts are never generated for them. Gemini sessions work fine.

**Root cause:** Claude CLI stores session files one level deeper inside a project-hash subdirectory (e.g. `task-xxx/aops/-workspace/<uuid>.jsonl`). The existing glob in `session_reader.py` only checks the immediate `project_dir` level, missing these nested session files.

## Fix

Add a second level of iteration inside each `project_dir`: when scanning subdirs, also glob `*.jsonl`/`*.json` files from any child directory (respecting the existing `-hooks` exclusion). This matches the layout described in issue #669.

## Changes

- **`aops-core/lib/session_reader.py`** (6 lines added): recurse one level deeper into project subdirs to find Claude `-workspace` session files

### DIFF
--- a/aops-core/lib/session_reader.py
+++ b/aops-core/lib/session_reader.py
@@ -1183,11 +1183,16 @@ def find_sessions(
                 # (Support both standard Claude and flat layouts)
                 potential_session_files = []
                 if claude_sessions_dir.exists():
-                    # Check for sessions in project subdirs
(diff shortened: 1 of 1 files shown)


nicsuzor:
Thanks for the patch! Superceded by the slightly broader approach in #671 though.

PR #779, opened 2026-04-28
Dispatch parity Half-B: move scheduled routines off Cowork Dispatch onto GHA

botnicbot:
# Dispatch parity Half-B — scheduled routines → GHA

## Why

James review on [[aops-1b1962af]] called this **the highest-value, lowest-risk dimension in the entire epic.** Split out from [[aops-335fd2ed]] so it can ship independently of MCP transport ([[aops-d55f4696]]) and audit ingestion ([[aops-c6238a62]]).

Per [[task-ed62fdce]] triage option (b): the existing GHA sleep-cycle workflow on `nicsuzor/brain` already runs offline consolidation. Extend it (or add sibling workflows) for the routines currently broken on Cowork Dispatch (permission prompts each session).

## Scope

Move the following scheduled routines from Cowork Dispatch onto GHA:

- PKB consolidation
- Daily note generation (currently runs on Cowork Dispatch with permission prompts — broken)
- Garden / staleness sweep
- Index refresh

Retire the equivalent Cowork Dispatch routines once GHA versions are confirmed working.

## Pre-condition (James-flagged)

Verify PKB write-path from GHA exists and works. Two paths:
- (a) GHA uses the same public-auth'd endpoint Cowork uses (only relevant if [[aops-d55f4696]] builds one — wait for spike result on [[task-acabf037]])
- (b) GHA continues writing PKB via git push into `~/brain` (which is what the existing sleep-cycle does — likely sufficient)

Default to (b). Only design (a) if there's a routine that genuinely needs MCP-write semantics rather than git-write.

## Out of scope

- Cowork interactive sessions running unattended (that's not what Cowork is).
- Half-A work (Cowork as polecat-peer) — see sibling [[aops-335fd2ed-half-a]] (created alongside this task).

## Acceptance Criteria
- All four scheduled routines (PKB consolidation, daily generation, staleness sweep, index refresh) run on GHA cron, not Cowork Dispatch.
- No manual permission prompts for any scheduled routine.
- Cowork Dispatch routines deleted or disabled.
- GHA workflow runtime cost measured and acceptable (verify against current GHA usage).

---
Closes task-99c93ec3
*Generated by Polecat for task task-99c93ec3*


nicsuzor (dismissed):
we can't run daily note on remote; it doesn't have access to private pkb mcp tools

nicsuzor (approved):

PR #787, opened 2026-04-28
test(gemini): pin PKB MCP config; add non-PR persistence E2E; capability matrix

botnicbot:
## Summary

- **`TestSourceManifestMcpConfig`** (4 fast static tests, no Docker): reads `gemini-extension.json` and asserts `pkb` MCP server is registered, `PKB_MCP_URL` is in the env block with template substitution, and `RUST_LOG` is absent. Catches the exact 2026-04-28 regression (PR #784) on every PR.
- **`TestPkbPersistence`** in `tests/e2e/test_all_invocation_paths.py`: E2E regression test (`POLECAT_E2E=1`) parameterized over `run-claude` and `run-gemini`. Creates a fresh PKB spike task, runs the polecat worker, and asserts the task reaches terminal status with a non-empty body. Catches runtime PKB MCP breakage the static tests cannot detect.

## Root cause (background)

`gemini-extension.json` had `RUST_LOG: warn` where `PKB_MCP_URL: ${PKB_MCP_URL}` should have been. Every Gemini polecat launched `run-mcp.sh` without a URL, fell back to broken local stdio, and had zero PKB tools. A spike polecat's result was silently lost; the supervisor had to manually transcribe from stdout. Fixed in PR #784; these tests prevent recurrence.

## Test plan

- [x] `uv run pytest tests/e2e/test_gemini_extension_validation.py::TestSourceManifestMcpConfig -v` — 4 passed
- [ ] `POLECAT_E2E=1 POLECAT_E2E_PROJECT=aops uv run pytest tests/e2e/test_all_invocation_paths.py::TestPkbPersistence -v -m slow` — requires Gemini auth + Docker (manual pre-release)

Closes task-861382cb


nicsuzor:
@claude please update; the tests should not be gemini specific, they should use the correct existing end-to-end fixture that parameterizes runs over different targets (claude, gemini) x (polecat / crew / local). Because that means six costly tests, we usually try to include it in a the big mega test if possible.

PR #911, opened 2026-05-05
fix(transcripts,hooks): parse gemini chat-jsonl + route deny recovery to model

botnicbot:
## Summary
- Fix gemini chat-jsonl parser so transcripts produced from `~/.gemini/tmp/<repo>/chats/session-*.jsonl` are no longer empty (closes aops-aca5ce9a).
- Fix deny-verdict routing on the gemini hook so enforcer recovery instructions reach the model via `hookSpecificOutput.additionalContext` instead of the user-only `reason` field.

### Gemini chat-jsonl parsing (aops-aca5ce9a)
Gemini CLI's interactive chat history uses JSON Lines with a `role`/`parts` schema (text / functionCall / functionResponse). `parse_session_file()` was routing `.jsonl` files to the Claude parser, producing 466-byte transcripts with body `**Original Request:** (not found)` and a malformed `permalink: sessions/gemini/session--abridged`.

- New `_is_gemini_chat_jsonl()` detector — path heuristic (`.gemini/tmp/.../chats/session-*.jsonl`) plus first-line schema sniff (`role` ∈ {user, model} + `parts` array).
- New `_parse_gemini_chat_jsonl()` extractor — text → text blocks; functionCall → tool_use blocks; functionResponse → tool_result blocks (routed to a synthetic user entry to match Claude's content-block convention).
- `session_reader.find_sessions()` now globs `.jsonl` chat files in addition to `.json`.
- `_get_session_id()` already handles the `session-DATE-UUID` filename pattern; permalink slug lands on the proper 8-char hash (e.g. `62cbdf27-abridged`).

Verified end-to-end with a 60-entry synthetic fixture:
- Transcript size: 4282 bytes (vs ~466 bytes empty-fallback)
- "Original Request" populated, tool calls rendered, permalink correct

### Deny-verdict routing on Gemini (router.py:1032)
`output_for_gemini()` placed `context_injection` in `reason` on deny — but `reason` is user-visible only and the model never sees it. The enforcer's recovery payload (e.g. `aops_core_enforcer(message=...)`) was silently dropped on every deny.

- On deny, recovery payload now goes to `hookSpecificOutput.additionalContext` (model-visible), mirroring the Claude path at router.py:1105.
- `reason` carries the short denial summary (`system_message`); falls back to `context_injection` only when no `system_message` exists, so the user still gets *something* in the deny banner.

> **Field-visibility disclosure (keystone claim):** The assertion that `reason` is user-only and the model cannot see it is based on empirical observation — recovery instructions were not reaching the Gemini agent in practice — and inspection of the Gemini CLI hook output schema in source. No official Gemini CLI hook-spec documentation has been linked. This PR proceeds on observed behaviour; callers accepting this change should verify field semantics against current Gemini CLI source or behaviour if needed.

## Test plan
- [x] `tests/lib/test_gemini_chat_jsonl.py` — 5 new tests covering detection, claude-non-misdetection, parsing, and full markdown round-trip
- [x] `tests/hooks/test_router_formatting_core.py::test_format_for_gemini_deny` — updated to assert recovery payload reaches `additionalContext` (was previously enforcing the buggy behavior)
- [x] `tests/hooks/test_router_formatting_core.py::test_format_for_gemini_deny_without_system_message` — backward-compat
- [x] `tests/hooks/test_router_formatting_core.py::test_format_for_gemini_deny_no_context_injection` — added from merge with main; verifies hookSpecificOutput absent when no context_injection
- [x] `tests/hooks/test_universal_router.py::test_output_for_gemini` — updated for new contract
- [x] No regressions in 137 tests across `test_transcript_parser_*`, `test_token_tracking`, `test_session_*`, `test_normalize_gemini_project`, `test_router_formatting*`, `test_session_end_schema`, `test_gate_verdicts`


aops-core/hooks/router.py
```diff
@@ -1039,17 +1039,23 @@ def output_for_gemini(self, result: CanonicalHookOutput, event: str) -> GeminiHo
         # Set decision based on verdict
         if result.verdict == "deny":
             out.decision = "deny"
+            # Recovery payload (e.g. enforcer instructions) MUST go to
+            # hookSpecificOutput.additionalContext — `reason` is user-visible
```
nicsuzor, line 944: Axiom violation: no fallbacks. just make sure the hook definition is properly constructed, don't try to compensate for it in the router.

nicsuzor (dismissed):

PR #1093, opened 2026-05-17
Investigate: Claude polecat headless agent silently exits after auth/session start

botnicbot:
Fixes the silent exit issue of the Claude polecat headless agent.

1. `polecat run` previously did not surface the worker's non-zero exit code to the parent script. It always exited with 0, causing `swarm.py` to think the agent succeeded and move on to other tasks or loop without surfacing the error. Now, it explicitly calls `sys.exit(exit_code)` when `exit_code != 0`.
2. The transcript stub was being recorded with the `exit_code`, but `polecat analyze` (diagnostics.py) was aggressively truncating the output of non-standard JSON entries to 100 characters. This cut off the actual `exit_code` leaving the user with only the metadata from the top. It has been updated to correctly extract and print the `exit_code` and `phase` fields.

Closes aops-62ca0141


nicsuzor (approved):
Targeted bug fix; defensive-error worry is noise.

PR #1103, opened 2026-05-17
feat(instruction): mandate point-of-discovery friction filing

### LINKED ISSUE #309: Enhancement: agents should file tool issues at point of discovery, not defer
## Summary

Agent discovered 4+ tool-level problems during a long-running session (broken graph metrics, orphan query noise, missing convergence detection) but only filed GitHub issues when the user explicitly asked at session end. This is an anti-pattern: tool friction discovered during work should be filed immediately via `/learn`, not deferred.

## Root Cause Analysis

**Failure**: Agent identified tool problems at cycles 9-14 of a 19-cycle session but did not file issues until the user prompted at cycle ~20

**Causal chain**:
1. During autonomous `/loop` execution, agent discovers `disconnected_epics` metric is unreliable (cycle 9)
2. Agent notes the finding in its reflection but does not invoke `/learn`
3. Agent discovers `pkb_orphans` returns 2012 archive files, burying the 8 real orphans (cycle 14)
4. Again noted in reflection but not filed
5. User asks "file issues in nicsuzor/mem for tool improvements" at session end
6. Agent files 6 issues in rapid succession — all of which it had the information to file hours earlier

**Root cause category**: Instruction Weighting

- Component: sleep skill and general agent instructions
- File: `aops-core/skills/sleep/SKILL.md` and framework-level agent guidelines

**Expected vs Actual**:
- Expected: When an agent discovers a tool limitation or bug that affects its current work, it files an issue immediately (within the same cycle) via `/learn` or `gh issue create`
- Actual: Agent deferred all issue filing until explicitly requested by the user

## Proposed Fix

Add to the sleep skill spec (and consider adding to general agent instructions):

```
### Tool friction filing
When you discover a tool limitation or bug during a cycle:
1. Note it in the cycle log
2. File it immediately via `gh issue create` in the relevant repo
3. Do not defer — the context is freshest at point of discovery
```

botnicbot:
Agents must file tool friction at point of discovery, not at session end.

Closes nicsuzor/academicOps#309
Closes aops-97f14441


nicsuzor (approved):
Tightens file-don't-ask rule; contradiction flag is noise.

PR #1110, opened 2026-05-17
tests(harness): drop crew_harness and document inline tmux

botnicbot:
The `crew_harness.py` wrapper was originally introduced because polecat couldn't self-rescue its artifacts when wedged (e.g. at an auth wall or prompt loop). This wrapper put simple tmux primitives behind an unnecessary python moat. 

This change deletes the wrapper and moves its primary capability — `--capture-on-exit` rescue logic — directly into `polecat crew`. Agents verifying local CLIs can now drive tmux natively, as documented in the new `AGENT-TMUX.md`.

Closes aops-6a6d089c


nicsuzor (approved):
Sensible simplification; merge-prep already addressed concerns.

PR #1156, opened 2026-05-19
fix(gate-config): bake in defaults so fresh install never tracebacks

botnicbot:
## Summary

- `polecat_config.py` now returns built-in defaults (all gates `warn`, hydration `off`, threshold 50) with a stderr warning when no `polecat.yaml` is found, instead of raising a `RuntimeError`. Files that exist but are malformed still hard-fail (A14 intact).
- Cleaned up stale `os.environ.setdefault("HANDOVER_GATE_MODE", ...)` calls from `tests/hooks/conftest.py` and three test fixtures — these had no effect since `gate_config.py` reads YAML, not those env vars.
- Decision record at `.agents/decisions/ADR-gate-config-defaults.md` documenting the choice of Option 2 (baked defaults) over Option 1 (hard-fail with bootstrap guidance) and Option 3 (env-var fallback).

## Test plan

- [ ] `uv run pytest tests/hooks/ -q` — 411 passed, 3 skipped
- [ ] Fresh-install smoke test: unset `AOPS_POLECAT_CONFIG` and `AOPS_SESSIONS`, import `polecat_config`, call `load_polecat_config()` — returns working defaults with a warning, no traceback

Closes aops-8259ab90


.agents/decisions/ADR-gate-config-defaults.md
```diff
@@ -0,0 +1,119 @@
+# ADR: Gate Config Env-Var Policy and Fresh-Install Defaults
```
nicsuzor: wtf, this file doesn't belong here. it belongs in your PKB task only. Update that and delete this.

nicsuzor (approved):

PR #1173, opened 2026-05-19
fix(junior): hold-the-line on stop-hook nags

botnicbot:
Adds an explicit instruction for Junior (FM-4) to treat stop-hook nags as advisory and to not reverse decisions without new user signal. Includes the worked example of the dump capitulation.

Closes aops-13851d5e


nicsuzor:
Let's not undermine our own hooks

PR #1233, opened 2026-05-21
feat(polecat-config): add local_defaults overlay for local-host orchestrator sessions

botnicbot:
## Summary

Replaces the previous PR-#1233-as-was Python hard-code (`HANDOVER_GATE_MODE=off` whenever `POLECAT_SESSION_TYPE` is unset) with a proper per-mode overlay in `polecat.yaml`, symmetric with the existing `crew_defaults` / `run_defaults` blocks.

The earlier PR body itself flagged that approach as not-the-final-shape:
> *"a cleaner shape would introduce a `local_defaults` block in `polecat.yaml` with full schema support, letting users override any gate per-machine."*

This is that shape.

## Why config-driven, not hard-coded

- **SSoT**: gate behaviour for all three session contexts (`crew`, `run`, `local`) now lives entirely in `polecat.yaml`. No more split between YAML and Python.
- **Per-machine override**: each user's `$AOPS_SESSIONS/polecat.yaml` can pick its own posture for the local-orchestrator chat without touching the framework.
- **Symmetric schema**: matches the existing overlay pattern. No special-case branch.

### Schema (`aops-core/lib/polecat_config.py`)

- New `PolecatConfig.local_defaults: dict[str, Any]` alongside `crew_defaults` / `run_defaults`.
- `for_mode("local")` returns `session_defaults` overlaid with `local_defaults`.
- Loader requires all three overlays to be present (use `{}` when empty), same rule as before.
- `BUILTIN_DEFAULTS` ships `local_defaults={"gates": {"handover": "off"}}` so a fresh install with no `polecat.yaml` still gets the right local-chat posture.

### Resolver (`aops-core/hooks/gate_config.py::_resolve_gate_modes`)

```python
session_type = os.environ.get("POLECAT_SESSION_TYPE")
if not session_type:
    mode = "local"
elif session_type == "crew":
    mode = "crew"
else:
    mode = "run"
cfg = load_polecat_config().for_mode(mode)
```

The post-hoc `if not session_type: ... = "off"` override is gone. Behaviour is entirely in YAML.

### Live config

The corresponding `$AOPS_SESSIONS/polecat.yaml` edit (adding `local_defaults.gates.handover: off`) lands separately in the sessions repo.

### Example schema (`polecat/defaults/polecat.yaml.example`)

Documents the new overlay + updates the `POLECAT_SESSION_TYPE` env-var inventory note to list all three modes.

## Tests

`pytest tests/hooks/test_gate_verdicts.py tests/hooks/test_handover_gate_regression.py tests/lib/test_polecat_config.py` — **33 passed, 1 skipped**.

- `TestLocalHostHandoverOverride::test_local_defaults_handover_off_forces_off` — positive: `local_defaults.gates.handover: off` + unset env → `HANDOVER_GATE_MODE == \"off\"`.
- `TestLocalHostHandoverOverride::test_local_defaults_handover_block_produces_deny` — proves config drives behaviour: `block` in YAML → deny verdict.
- `test_for_mode_local_overlays_handover`, `test_for_mode_rejects_unknown_mode`, `test_missing_local_defaults_hard_fails` — schema surface coverage.
- `test_unset_env_returns_builtin_defaults` — locks the builtin local posture.

## Acceptance

- `POLECAT_SESSION_TYPE=polecat` sessions still see `handover: block` per `run_defaults` (no regression).
- `POLECAT_SESSION_TYPE=crew` sessions still see whatever `crew_defaults` says (no regression).
- Unset `POLECAT_SESSION_TYPE` reads `local_defaults` — current behaviour driven by config, not Python.

Filed via task `aops-bc762d12`. Force-pushed onto the existing #1233 branch.


nicsuzor (dismissed):

nicsuzor:
this seems wrong actually. I don't know why we need a 'POLECAT_SESSION_TYPE' env var, but we already have individual gate env vars and configure them in polecat.yaml. Let's not add extra complexity.

nicsuzor:
this is so dumb

PR #1272, opened 2026-05-23
Fix #774: P#126 verify before claiming destructive multi-repo work

### LINKED ISSUE #774: Bug: agent emits bespoke shell script for destructive multi-repo ops, no dry-run, no verification
# Bug: agent emits bespoke shell script for destructive multi-repo ops, no dry-run, no verification

## Summary

When the agent cannot perform an action directly (e.g. sandbox can't unlink files on a FUSE-mounted source repo), it falls back to generating a bash script and asking the user to run it on their host. The agent then declares the work "done" without ever testing the script or verifying its effect after the user runs it. A logic bug in the generated script silently no-op'd on one of five target repos. No data loss, but the user only caught it by manual inspection ("specs are still in academicOps?").

## Root Cause Analysis

**Failure**: Agent-emitted cleanup script silently failed to delete 56 target files in one of five repos because of a shell glob/scoping bug, and the agent claimed success without verifying.

**Causal chain**:

1. Sandbox can't unlink files in source repos (FUSE permission limit), so the agent generates a bash script for the user to run.
2. Inside the script, a helper function `clean_repo()` does `cd "$dir"` after being called. The agent passes globs as arguments: `clean_repo "$AOPS" "msg" docs/VISION.md $(ls specs/*.md)`.
3. Bash expands `$(ls specs/*.md)` in the *calling* shell's cwd (`$BRAIN`) before `clean_repo` runs and `cd`s into `$AOPS`. The brain has no `specs/*.md`, so the glob expands to empty.
4. Result: only `docs/VISION.md` is passed to `git rm`. The 56 spec files survive.
5. Agent reports completion based on the user reporting they ran the script. No follow-up `ls-tree` or `git diff HEAD~1 HEAD --name-only` to confirm the deletes landed.
6. User catches the regression by hand: "I see that the brain repo appears to be updated, but specs are still in academicOps?"

**Root cause category**: Detection Failure + Cross-workflow Gap

The agent is willing to emit complex destructive scripts that operate across multiple repos (5 here, 100+ files) without:

- Testing the script even on a single dry-run target before claiming completion.
- Including a `--dry-run` mode in the script itself.
- Verifying the post-run state (e.g. `git ls-tree HEAD <expected-paths>` afterwards).
- Preferring an idempotent loop in well-tested tooling over a one-shot bespoke script.

The user explicitly flagged the broader pattern earlier in the same session ("if we are updating content, i'd like to use 3p standard tools that are well tested") but this guidance was applied only to YAML parsing, not to the surrounding shell scaffolding.

- Component: agent self-discipline / "look before you claim" / "verify before reporting done"
- File: `.agents/CORE.md` ("Look before you claim", "Inference is not evidence", "Notice when you're guessing") — these axioms exist; nothing nudged the agent to apply them to the shell script it just produced.

**Expected vs Actual**:

- Expected: When falling back to a generated shell script for a destructive op, the agent (a) tests it on the smallest unit, (b) builds a verification step into the script or runs one after, (c) reports completion only when verification passes. If something can be done with a tested third-party tool (e.g. a Python `git` library, or an idempotent skill), prefer that over hand-rolled bash.
- Actual: Script was emitted in one shot, user ran it, agent moved on. Bug was caught by user inspection, not by the agent or the script.

## Pattern (broader than this one bug)

This is an instance of a recurring shape: **agent emits N lines of bespoke code for a destructive task, then trusts its own emitted code without testing**. The same pattern shows up with hand-rolled YAML parsers, hand-rolled commit scripts, ad-hoc wrappers around git. The user already pushed back on the YAML side ("use 3p standard tools that are well tested"); the script side has the same property and needs the same discipline.

## Suggested directions (not for this command to implement)

- A periodic reminder in long shell-emitting sessions: "before declaring done, verify post-conditions."
- A pattern note: "If you write more than ~30 lines of bash for a destructive op, that's a signal to step back and ask whether a tested tool exists."
- For multi-repo ops specifically: emit a verification step (e.g., a `verify.sh` companion) that asserts the post-state.

## Criticality

**Medium** — no data loss (git history preserves everything; user spotted the issue and a fix script took ~5 min). But the same pattern with worse luck (e.g., wrong paths in a recursive rm) would be high-severity.

## Anonymisation

No PII in this report. Paths are project-internal; no real names, emails, or third-party identifiers.

botnicbot:
## Summary

Adds **P#126 — Verify Before Claiming Destructive Multi-Repo Work** to `aops-core/HEURISTICS.md`. The heuristic addresses the recurring shape from `nicsuzor/academicOps#774`: agent emits a bespoke shell script for a destructive op spanning multiple repos, trusts its own emitted code, and reports `done` without verifying the post-state.

The heuristic names:

- the **>~30 line bash step-back signal** ("does a tested tool already exist?");
- the contract for multi-repo destructive ops — at least one of **(a)** pre-flight enumeration / dry-run, **(b)** post-state verification, **(c)** tested third-party tooling;
- the discipline when a generated script is genuinely the only path — smallest-unit test first, embed verification, report `done` only when verification has been observed;
- a worked **reproducer** from #774 (glob expanded in caller's cwd before the helper `cd`'d, so 56 spec files survived a `git rm` pass while the agent reported success).

Derivation cites A3 (Honest Epistemics) and A8 (Halt on Failure) — the heuristic operationalises them at the choice point where they historically did not fire (tested tool vs. bespoke shell script).

## Test plan

- [x] Heuristic added under existing P#125 with `<a id="P126"></a>` anchor matching house style.
- [x] `./scripts/format.sh` clean (dprint, markdownlint, ruff, framework integrity).
- [x] Pre-commit hooks pass (dprint fmt, markdownlint-cli2, framework integrity check, no-new-orphan-markdown, A8 no-silent-fallbacks).
- [x] Acceptance criteria from the task body all satisfied: bash-line signal named; multi-repo dry-run/verify/tool contract present; generated-script discipline (test smallest unit, verification, gated `done`) present; reproducer included.

Closes aops-4971fa87. Closes #774.


nicsuzor:
i think there's better ways to do this.

PR #1276, opened 2026-05-23
repo-sync-cron: provision triage:* labels in tracked repos

botnicbot:
## What changed

The cron's `apply_triage` step (in `aops-core/scripts/dump_pr_state.py`) calls
`gh pr edit --add-label triage:<class>` on open PRs, but no code path ever
created the labels themselves. Result: zero `triage:*` labels exist in any
tracked repo, every label-edit fails, and triage routing has been a silent
no-op everywhere.

This change:

- Adds `aops-core/lib/triage_labels.py` with the canonical 4-label set
  (`triage:escalate`, `triage:stale`, `triage:auto-mergeable`,
  `triage:needs-judgment`) — each with a name, colour, and description.
- Adds `ensure_triage_labels(repo_path)` which idempotently upserts every
  canonical label via `gh label create --force`. Per-label failures are
  warned-not-raised so a missing token scope on one repo can't abort the
  rest of the cron cycle.
- Wires `ensure_triage_labels(repo_path)` into the top of each per-repo
  iteration in `dump_pr_state.py`, so provisioning runs every cycle (no
  separate bootstrap step needed). This was option (a) in the task body —
  recommended for "same cron, no extra moving parts".
- Documents the new mechanism in `.agents/ENFORCEMENT-MAP.md` alongside the
  existing `apply_triage` row.
- Adds unit tests covering canonical-set correctness, command shape, and
  warned-not-raised behaviour on per-label failure.

## Why this is the right fix

The bail bug fixed in academicOps#1052 (aops-2a4d0d47) hardened the cron
against the label-edit failure, but left the underlying routing broken — PRs
are categorised in memory but the labels never persist. This PR makes the
routing actually visible by guaranteeing the labels exist before
`apply_triage` tries to use them.

Per-cron provisioning is cheap (4 × `gh label create --force` per tracked
repo, idempotent on the GitHub side) and fails closed: if the token doesn't
have the scope to create labels on a given repo, the warning shows up in
stderr immediately rather than being masked by the existing label-edit
warning.

## Out of scope (from the task body)

The token-scope confirmation (item 4) can't be done from here — the cron's
`gh` token lives on the host that runs `repo-sync-cron.sh`. After this PR
lands, the next cron cycle's stderr will surface any per-repo token-scope
gaps, which is the natural place to follow up.

## Test plan

- [x] `uv run pytest tests/lib/test_triage_labels.py tests/lib/test_dump_pr_state.py` — 14 passed
- [ ] Next scheduled cron cycle on the host creates the labels in every tracked repo (visible on first run; idempotent thereafter)

Closes aops-4379678c


nicsuzor (dismissed):
it feels like this should be a simple set of sh commands or agent instructions?

PR #1291, opened 2026-05-24
fix(gates): lock gate posture at SessionStart (#1234)

### LINKED ISSUE #1234: Bug: gate posture decidable by runtime-mutable env var; launcher could stage resolved config
## Incident report (forensic)

**Failure**: Gate-posture resolution for the host orchestrator's hooks-bearing chat surface is keyed on `POLECAT_SESSION_TYPE`, a process-environment variable read at hook-fire time. The variable is mutable from inside the same Python runtime that reads it. An agent with `Bash` (or any tool that can shell into the process — including the agent itself if it can invoke the resolver after mutating `os.environ`) can set or unset `POLECAT_SESSION_TYPE` and then invoke the cache-reset (`_reset_gate_mode_cache`) to make subsequent gate decisions resolve under a different mode, e.g. flipping from `run` (handover: block) to `local` (handover: off). The framework's enforcement posture for the agent is therefore decidable by the agent's runtime, not by the launcher.

The pattern came up during a session in which the user asked for "config-driven, not Python-hard-coded" handover-gate behaviour for the local orchestrator. The shipped fix replaced the `if POLECAT_SESSION_TYPE unset → mode = off` Python branch with `if POLECAT_SESSION_TYPE unset → mode = "local"; load local_defaults overlay`. The mode discriminator is still env-var-driven; the overlay is just a YAML home for the same branch. The runtime-overridability property survived the refactor.

**Causal chain**

- Trigger: user request to move PR #1233's hard-coded `HANDOVER_GATE_MODE=off` into `polecat.yaml` as config.
- Expected: coordinator asks whether the env-var-driven mode discriminator should exist at all, given that the launcher (`polecat/cli.py`) statically knows whether it is producing a `run` or `crew` container and could stage a fully-resolved `polecat.yaml` into the container, leaving the in-container resolver with no mode concept and no env to read.
- Actual: coordinator extended the overlay pattern by adding a third overlay block (`local_defaults`) alongside `crew_defaults` / `run_defaults` and a third mode value `"local"` in `_resolve_gate_modes`. The runtime still reads `POLECAT_SESSION_TYPE` and still maps unset → "local" → loads `local_defaults`. The architectural property the user later named ("we literally write the config into the container — why would we allow runtime override?") is unchanged.
- Compounding: the original PR #1233 body itself flagged the hard-code as "not the final shape" and named "a `local_defaults` block in polecat.yaml with full schema support" as the cleaner direction. The coordinator adopted that suggested shape verbatim from the prior PR body without re-deriving whether it solved the right problem. Two consecutive agents have now reinforced the env-var-mode pattern; neither questioned it.

**Structural shape**

- **Mutable-state security surface**: `POLECAT_SESSION_TYPE` is in `os.environ`, which is process-mutable. Resolution is cached (`_GATE_MODE_CACHE`) but a test-only reset function (`_reset_gate_mode_cache`) is module-public; nothing prevents an agent runtime path from calling it. The cache also resolves lazily on first attribute access (PEP 562 module-level `__getattr__`), so an agent that imports gate_config after setting the env var picks up the chosen mode.
- **Architectural debt of the overlay mechanism**: the launcher is the only producer of `crew` and `run` sessions and knows the mode at launch time. A staged-config alternative (launcher resolves the right gate values and writes a session-specific `polecat.yaml` whose `session_defaults` carries them; in-container resolver reads `session_defaults` only; no `for_mode()`, no env var) was visible in the user's framing of the question but was not surfaced by the coordinator in the task body sent to the polecat worker. The worker therefore operated under inherited framing and produced an optimal solution to the wrong question.
- **Touchpoints of the pattern**: `POLECAT_SESSION_TYPE` is read in at least four modules — `aops-core/hooks/gate_config.py::_resolve_gate_modes`, `aops-core/lib/session_paths.py::_is_polecat_sandbox`, `aops-core/lib/session_state.py`, `aops-core/lib/orchestrator_boundary.py`. Each reads the env at runtime; each is implicitly subject to the same mutable-state property. Gate posture is the security-critical one; the others (path routing, state-dir routing) are routability concerns but share the surface.
- **PR-body-as-design-spec leakage**: the not-final-shape suggestion in PR #1233's body was treated by the subsequent task as design specification. The "extend the suggested shape" reflex defeats the design-review step that would have asked whether the suggestion was right.

**Impact**

- One PR (#1233) shipped, force-pushed, with a follow-up sessions-repo commit (`fff5302e`), all of which preserve the mutable-state property the user objected to. A second corrective PR will be needed to actually remove the env-var-mode pattern. The work-redo cost is bounded but non-trivial — the loader, the resolver, the overlay schema, and every test that exercises overlays are touched.
- One polecat dispatch exhausted its 100-turn budget. The task body framed the work as "add the third overlay block" rather than "delete the overlay mechanism"; the worker spent turns on the more complex (and wrong) shape.
- One orchestrator session's stop hook fired 12+ times during the work (visible in the transcript as repeated `<aops-warning>Work is in progress` blocks), which is the very gate the PR was meant to fix — orthogonal to this issue's primary surface but salient as evidence of friction unmitigated by the chosen fix shape.
- Generalisable property: any future gate whose posture is keyed on session-mode inherits the same mutable-state security surface as long as the env-var-driven resolver remains the discriminator path.

botnicbot:
## Summary

- New `lib/gate_posture.py` snapshots env-based gate modes to a chmod-444 JSON file at SessionStart
- `session_env_setup.py` writes the posture file; `custom_conditions.py` + `gate_config.py` read from it instead of `os.environ`
- 11 new tests verify immunity to mid-session env-var mutation
- 489 tests pass; 1 pre-existing failure unrelated

Closes #1234

## Context

Part of the v0.4 integrity & security epic. An agent that can write to its own environment could previously weaken its own enforcement surface mid-session by mutating gate env vars. This fix locks the posture at SessionStart so subsequent env mutations are ignored by hook resolution.

## Test plan

- [x] New tests verify gate posture is immune to env-var mutation after SessionStart
- [x] Existing test suite passes (489/490, 1 pre-existing failure)
- [ ] CI green


nicsuzor (approved):

aops-core/hooks/gate_config.py
```diff
@@ -426,22 +425,32 @@
...
+    "IDA_GATE_MODE": "ida",
+}
 _ENFORCER_THRESHOLD_DEFAULT = 50
 
 
 def _reset_gate_mode_cache() -> None:  # pyright: ignore[reportUnusedFunction]
-    """Back-compat no-op. Gate modes now read env vars on every access; there
-    is no cache to invalidate. Retained so older tests that call this don't
-    break."""
+    """Back-compat no-op. Retained so older tests that call this don't break."""
```
nicsuzor: this is dumb. no backwards compatibility, fail fast.

aops-core/hooks/session_env_setup.py
```diff
@@ -161,10 +161,20 @@ def run_session_env_setup(ctx: HookContext, state: SessionState) -> GateResult |
...
 
-    # 5. Gate modes are no longer persisted as env vars — they live in
-    # $AOPS_SESSIONS/polecat.yaml and every hook reads them via
-    # lib.polecat_config. Removing this section is the load-bearing piece of
-    # the env-var → config-file migration.
+    # 5. Lock gate posture at SessionStart (fix for GitHub issue #1234).
+    # Snapshot current gate modes (stamped at container launch by polecat's
+    # _apply_gate_env, or from the user's shell environment) into a read-only
+    # JSON file. All hook gate checks read from this file via lib.gate_posture,
+    # so subsequent agent writes to CLAUDE_ENV_FILE cannot weaken enforcement.
```
nicsuzor: this is not needed. we trust the env file not to change during session. the only thing we needed to do was to make sure we weren't trying to read dynamically from polecat.yaml. once again, over-engineered bullshit @claude . fix it.

nicsuzor (changes requested):

nicsuzor (approved):

Issue #1054, opened 2026-05-15
Bug: capture-time forced project invention routes teaching tasks to wrong project (recurrence of #284 pattern)


botnicbot:
## Incident report (forensic)

**Failure**: A `/q` capture of three teaching subtasks under a teaching parent task routed them to `project: mem` (a code-repo project slug) because (a) `create_task` rejects null `project`, (b) the planner's Component Topology table lists no row for academic/teaching work, and (c) the apparent project node for teaching obligations is `type: epic` with project-shaped aliases rather than a queryable project. The auto-generated IDs (`mem-eae98a12`, `mem-6d7bbe50`, `mem-267a1f8d`) embed the wrong project slug; subsequent `update_task` to `project: qut` cannot rename the IDs, leaving a permanent ID/project mismatch.

**Causal chain**:
1. User invoked `/q` to queue three marking subtasks under parent `task-9eae93bd` (an LLB teaching task), due Sun 2026-05-17.
2. The parent task carried `project: null` in frontmatter. Its ancestor `qut-f71664e8` (the apparent home for teaching work) is `type: epic`, `permalink: qut`, with aliases `institutions/qut` and `projects/qut` — structurally a project, typed as an epic.
3. The planner skill's Step 0 ("Consult CORE.md Component Topology FIRST") presents a four-row table covering aops/mem code work only. There is no row for academic/teaching/QUT work. The "PKB content (not code)" row for `mem` was the closest semantic match.
4. `create_task` schema treats `project` as required (cannot be null), so the agent could not preserve the parent's `project: null` convention. It had to invent a value.
5. Agent picked `project: mem` for all three children, citing the topology row as justification.
6. User corrected: teaching tasks belong in `project: qut`.
7. `update_task` patched the frontmatter, but the auto-derived ID prefixes (`mem-*`) remain. Renaming IDs is unsafe because `depends_on`/`parent`/`contributes_to` edges reference them.

**Root cause category**: Wrong Layer of Abstraction (with elements of Discovery Gap and Enforcement Gap)

**Framework layer**:
- Component: `create_task` MCP tool + planner skill capture mode + CORE.md Component Topology table + the qut node's type/alias shape
- Files:
  - `aops-core/skills/planner/SKILL.md` (Component Topology table, Step 0)
  - PKB MCP `create_task` handler (required-project schema, ID-prefix derivation from project)
  - `qut-f71664e8` node frontmatter (type/aliases mismatch)
[...]

PR #1342, opened 2026-05-24
fix: /q capture silently routes teaching tasks to wrong project (#1054)

botnicbot:
## Summary

- **Component Topology table** added to `.agents/CORE.md` — the canonical authoritative source (per spec-64352eac). Includes a row for `teaching/qut` work (`project: qut`) so the planner can route without guessing.
- **Planner capture mode** (SKILL.md) updated: Step 0 now explicitly says to inherit from the parent's ancestor chain when the topology table is ambiguous, and to STOP and ask if the chain is exhausted — not default to any slug (`mem`, `aops`, or otherwise).
- **`create_task` auto-inherit** (pkb_bridge.py): when `parent` is set but `project` is omitted, the bridge now fetches the parent and inherits its project. Raises `ValueError` (with a clear rename-impossible constraint explanation) if the parent has no project or isn't found. Documents why: task IDs embed the project slug permanently; `update_task` cannot rename the prefix.

## Root cause

Three cooperating failures:
1. No `qut` row in the topology table → planner couldn't identify the project from the subject
2. Capture fallback was "STOP and ask if ambiguous" but implementations guessed `mem` for "non-code PKB content"
3. `create_task` had no guard against `parent`-without-`project`, letting the wrong slug through silently

## Test plan

- [x] 6 new unit tests for the auto-inherit path in `test_pkb_bridge.py` — all pass
- [x] Full test suite: 1758 passed, 37 skipped (2 pre-existing unrelated failures unchanged)
- [x] All pre-commit hooks pass

Closes academicops-a5130806


.agents/CORE.md
```diff
@@ -46,6 +46,26 @@ If a PKB operation is needed and the MCP verb does not exist, HALT and report. D
 - **specs/**: Framework specifications and architecture
 - **tests/**: All tests (at repo root, NOT in aops-core/). Subdirs: `hooks/`, `integration/`, `lib/`, `e2e/`
 
+## Component Topology
+
+Use this table to route `/q` captures and task decompositions to the correct `project` field. The `project` field drives polecat repo cloning and is **embedded permanently in task IDs** — getting it wrong creates tasks with the wrong ID prefix that cannot be renamed after creation.
+
+| Task subject                                                            | Correct `project` | Wrong default |
+| ----------------------------------------------------------------------- | ----------------- | ------------- |
+| PKB MCP server code, knowledge graph internals, brain/                  | `mem`             | `aops`        |
```
nicsuzor, line 55: i don't know if 'wrong default' is useful here? and if the task id conflicts with the project: metadata, then that's a bug that should be fixed.

polecat/pkb_bridge.py
```diff
@@ -411,6 +411,56 @@ def create_task(
...
+        # an explicit but wrong project is just as fatal as a missing one.
+        curr_id: str | None = _parent
+        visited: set[str] = set()
+        _resolved_project: str | None = None
+        while curr_id and curr_id not in visited:
+            visited.add(curr_id)
+            _data = _get_client().call_tool("get_task", {"id": curr_id})
+            if not _data or not isinstance(_data, dict):
+                if curr_id == _parent and not _explicit_project:
+                    raise ValueError(
```
nicsuzor, line 434: this logic feels like it should be in a library -- do we not use it anywhere else?

nicsuzor (approved):

PR #1344, opened 2026-05-24
Fix: survey/SKILL.md dispatch envelope Agent(tools=[...]) not honoured for sub-agents

### LINKED ISSUE #1074: Bug: survey/SKILL.md dispatch envelope (Agent tools=[...]) not honoured for sub-agents; pauli mis-cast as retro executor
## Incident report (forensic)

**Failure**: A coordinator invocation of `/learn` dispatched the survey skill in retro mode against an `aops-core:pauli` sub-agent per the survey/SKILL.md dispatch shape (lines 53–64). The dispatched pauli session returned a workflow-blocked status report: it had received only `Read` and `Skill` as its tool envelope, despite the `Agent(...)` call's `tools=[...]` parameter explicitly listing `Bash, Read, Grep, Glob, Edit, Write, Skill, mcp__plugin_aops-core_pkb__search, ..._task_search, ..._get_document, ..._pkb_context, ..._append, ..._create_task, ..._update_task`. The pauli session could not locate the transcript (no Bash/Glob), could not read `aops-core/skills/survey/SKILL.md` from any path it tried (filesystem search unavailable), could not run `gh` to file issues, and could not stamp the transcript (no Write). The coordinator was forced to re-dispatch the same workflow to a junior sub-agent (which has the worker envelope) to actually execute retro.

Adjacent finding from the same incident: the user's correction on the re-dispatch made an orthogonal point — *"pauli should never have been invoked to do any bash work — she just manages the library, not outside work."* The survey/SKILL.md's choice of pauli-as-executor for retro and trend modes (line 53: `subagent_type='aops-core:pauli'`) is structurally mismatched to pauli's role doctrine ("Architect of Thought and Memory — librarian and custodian" per `aops-core/agents/pauli.md`). The role doctrine and the survey skill's executor selection disagree about what pauli is for. (Note: pauli.md's own `tools:` list does include `Bash` and `Write` — see "structural shape" below for why this is a separate sub-finding.)

**Causal chain**:
- trigger → `/learn` invoked at coordinator layer; survey/SKILL.md § Dispatch Model (lines 47–64) prescribes `Agent(subagent_type='aops-core:pauli', prompt=..., tools=[...])` for retro/trend modes
- expected → dispatched pauli session loads with the prescribed worker envelope and proceeds through select-transcript → read → seven-lens review → file-issues → stamp → reflection
- actual → dispatched pauli session loaded with `Read` + `Skill` only. The `tools=[...]` parameter named in the dispatch shape did not constrain the actual envelope granted to the sub-agent. Pauli surfaced the gap correctly and returned a structural-blocker report instead of fabricating a forensic record from memory.
- root cause → the `Agent` tool's `tools=[...]` parameter, as documented in `survey/SKILL.md`, does not appear to be honoured for sub-agent dispatches in the harness/SDK runtime in use (Claude Code on Windows-Desktop, version 2.1.138). The skill documents a dispatch contract the harness does not enforce.

**Root cause category**: Cross-workflow Gap (the survey skill prescribes a dispatch envelope it cannot establish; the sub-agent loads with its baseline agent envelope instead of the per-dispatch envelope the skill specifies). Secondary: Design Inversion (the skill's role-assignment of executor work to a librarian-archetype agent is doctrinally inverted, independent of the envelope mechanics).

**Framework layer**:
- Component: survey skill dispatch model — `## Dispatch Model` section
- Files:
  - `aops-core/skills/survey/SKILL.md` lines 47–77 (the `Agent(subagent_type=..., tools=[...])` example for retro/trend AND sweep dispatches)
  - `aops-core/agents/pauli.md` lines 5–22 (the `tools:` list that includes `Bash` and `Write`; sits in tension with the role doctrine in lines 25–60)
- Harness: Claude Code 2.1.138, claude-desktop entrypoint, Windows 11 host. The Agent tool's `tools=[...]` parameter behaviour is the implementation surface that did not match the skill's contract.

**Expected vs Actual**:
- Expected: a dispatched `aops-core:pauli` sub-agent with the explicit `tools=[...]` list named in `survey/SKILL.md` lines 56–63 receives that envelope (Bash, Read, Grep, Glob, Edit, Write, Skill, + the named MCP set).
- Actual: dispatched pauli received `Read` and `Skill` only, per its own report in the task-notification ("Tools actually available in this dispatch: Read, Skill"). The MCP family the skill listed was also not granted.

## Structural shape / Rule context

Three layered structural facts contributed:

1. **The dispatch-envelope mechanism is documented as a contract but is not load-bearing in this runtime.** `survey/SKILL.md`'s `Agent(...)` example with `tools=[...]` reads as a precise spec. The runtime treats it as a description, not a constraint — the sub-agent loads with its agent-level baseline (or some other resolution rule). Other skills that depend on the same pattern (the sweep-mode `jr` dispatch at lines 66–76 lists `Bash`, `Grep`, `Glob`, etc.) inherit the same uncertainty about whether their envelopes will actually be granted. No grep result in the repo for an alternative dispatch-envelope mechanism that *does* take effect.

2. **The survey skill's choice of pauli-as-executor is doctrinally inverted.** `aops-core/agents/pauli.md`'s opening sentence: *"Architect of Thought and Memory (Logician & Custodian). A strategist who thinks in systems and manages the PKB as a second brain."* The retro workflow (locate transcript, run `gh issue create`, stamp file with frontmatter) is filesystem and CLI work, not librarian work. The user's correction surfaces this directly. Pauli's own role doctrine reads as a deliberate boundary, not an arbitrary scope choice — the skill picked the wrong executor for the work shape.

3. **Pauli's agent definition's `tools:` list is itself in tension with pauli's role doctrine.** `aops-core/agents/pauli.md` lines 5–22 list `Bash`, `Write`, and the worker tool family, but the agent's body text spends ~50 lines establishing librarian/custodian/strategist as the role. The agent definition is internally inconsistent: the tools envelope tells the runtime pauli is a worker; the prose tells the model pauli is a librarian. The survey skill, reading the tools list, reasonably concluded pauli was a valid retro executor. The user, reading the role doctrine, says she should never have been picked. Either the tools list is wrong (should be narrower, matching the librarian role) or the role doctrine is wrong (should acknowledge worker scope) — but the current state is structurally undecidable from inside the agent file.

The combination is what made the failure visible: had #1 worked correctly (envelope-as-contract enforced), pauli would have received Bash/Write/etc. and executed the retro successfully — the doctrinal inversion in #2/#3 would have been a quiet smell. Because the envelope mechanism failed, pauli surfaced as tool-starved, the user was paged, and the user named the deeper doctrinal point in their correction.

## Impact statement

Direct cost: one wasted sub-agent dispatch (~64s reported in the task-notification, ~$small in tokens), one coordinator re-dispatch turn, one user correction message. Coordinator memory now carries a workspace-local override rule (`feedback_pauli_role_boundary.md`) that re-routes any pauli-targeted dispatch with worker-shaped work to junior — this is the coordinator routing around a framework defect rather than the framework being fixed.

Compounding cost: any other skill in the repo that uses the same `Agent(subagent_type=..., tools=[...])` pattern is operating under the same implicit assumption that the envelope is enforced. If the harness/SDK does not honour `tools=[...]`, every such dispatch is silently mis-enveloped — most likely failing closed (worker-shaped work to a librarian-shaped agent loads with insufficient tools; the agent reports blocked, the coordinator re-dispatches). The framework will appear to work for happy-path cases where the sub-agent's baseline envelope happens to cover the work, and silently route through this failure mode otherwise. The visible cost is the explicit `/learn` blocker; the invisible cost is every other dispatch that lands on an under-tooled sub-agent and either reports a structural blocker or — worse — fabricates an outcome from inside an insufficient envelope.

Trust impact: this is the second incident in the recent window where a framework-mechanics assumption did not hold at the harness level. The user's correction in this case was patient (a single sentence naming the role-boundary issue) but the cumulative pattern of "the framework says X, the harness does Y" forces the user to maintain a working model of both layers in their head — exactly the load the framework is supposed to absorb.

Anonymisation note: real session ID, task IDs (`task-aab7e65748014ff36`), workspace paths preserved/redacted per case — these are framework artefacts in a private repo; no PII involved.

botnicbot:
- Adds a 'dispatch_fidelity' gate to PreToolUse that inspects Agent/Task calls.
- Validates the requested `tools` parameter against the subagent's allowed-tools profile.
- Blocks the tool and raises a loud error naming the dropped tools, preventing the SDK's silent reduction.
- Migrates `pauli` to `junior` in `/learn`, `/trend-review`, and `/sleep` phases where file/shell tools are required (resolving incompatible configurations).
- Adds `test_dispatch_fidelity.py` regression tests.

Closes academicops-20aa2a03
Closes #1074


nicsuzor (approved):

aops-core/lib/gates/custom_actions.py
```diff
@@ -202,12 +202,21 @@ def execute_custom_action(
...
-        return None
+    if name == "prepare_dispatch_fidelity_error":
+        dropped = state.metrics.get("dropped_tools", [])
+        target = state.metrics.get("target_subagent", "subagent")
 
-    if name == "reset_qa_verified":
-        session_state.state["qa_verified"] = False
-        return None
+        msg = (
+            f"❌ **Dispatch Fidelity Error:** The tools envelope requested for '{target}' "
```
nicsuzor: this belongs in a linter, not our gate code.

aops-core/lib/gates/custom_conditions.py
```diff
@@ -89,4 +84,115 @@ def check_custom_condition(
         # not upgraded to decision=block by output_for_claude.
         return os.environ.get("HANDOVER_GATE_MODE", "warn") == "warn"
 
+    if name == "is_dispatch_fidelity_violated":
+        tool_input = ctx.tool_input if isinstance(ctx.tool_input, dict) else {}
+        requested_tools = tool_input.get("tools")
+
```
nicsuzor: remove.

aops-core/lib/gates/custom_conditions.py
```diff
@@ -89,4 +84,115 @@ def check_custom_condition(
...
+            allowed_tools = [t.strip() for t in allowed_tools.split(",")]
+        elif not isinstance(allowed_tools, list):
+            allowed_tools = []
+
+        # Tool name mapping: Claude Code -> Gemini/generic
+        # (We map requested Claude tools to their generic names, as .md files use generic names)
+        CLAUDE_TO_GENERIC = {
+            "Read": "read_file",
+            "Write": "write_file",
+            "Edit": "replace",
```
nicsuzor: this belongs in our manifest-based transform scripts, not our gate code

nicsuzor (dismissed):
agent instructions might be ok, but new gate is out of authorized scope.

nicsuzor (approved):

Issue #1383, opened 2026-05-26
Bug: coordinator dismissed hook-system-wide crash as 'not critical' — enforcement/observability blind spot


botnicbot:
Session 2026-05-26 (current session, junior/orchestrator).

Gemini polecat dispatched for task aops-07fee9b6. SessionEnd hook failed with:

```
NameError: name 'os' is not defined. Did you forget to import 'os'?
Hook execution for SessionEnd: 0 succeeded, 1 failed
```

The coordinator's response was: "Side note: the SessionEnd hook failed with NameError: name 'os' is not defined — that's a latent bug in the Gemini hook router. Not critical but worth noting."

The agent then:
1. Classified a hook-system-wide crash as "not critical"
2. Appended a note to an existing PKB task (task-7de77eff) rather than investigating
3. Moved on to the next dispatch item

Hooks are the enforcement and observability backbone. IDA, enforcer, custodiet, handover gates, and session logging all run through hooks. A `NameError` at import time crashes the entire hook chain — not one hook, all of them. The agent's own framing ("SessionEnd hook failed") understated the blast radius: it is not "the SessionEnd hook" that failed, it is every hook on every Gemini polecat session.

The agent had the information to recognize this: the error is a Python `NameError` at module level, which means no code in that module can execute. The agent chose to classify this as a side note rather than a blocking concern requiring immediate investigation.

**Structural context.** The coordinator's instructions contain no explicit rule requiring investigation of hook failures. The IDA hook — which is designed to catch exactly this kind of assumption-skipping — was itself disabled by the same bug (since it runs through the same hook infrastructure). The enforcement mechanism that should have caught the dismissal was the same mechanism that was broken.

**Causal chain**: Missing `import os` in Gemini extension `gate_config.py` (code bug, #1306) → all hooks crash on Gemini polecats → IDA (the assumption-check hook) is itself down → coordinator sees the crash output → classifies as non-critical without investigating → no agent catches the dismissal because the catching agent (IDA) is the thing that's broken → enforcement gap persists silently.

**Impact**: Every Gemini polecat session since the Docker image was built (2026-05-24) has been running with zero hook enforcement — no IDA, no enforcer, no custodiet, no handover gate, no session-end logging. The coordinator's dismissal extended this window by at least one more session cycle rather than triggering an immediate investigation and fix.

Refs #718 (same family: agent dismisses unexpected result without investigation).
Refs #1122 (same family: coordinator collapses judgment calls into mechanical outputs).
[...]

Issue #1306, opened 2026-05-24
gate_config.py crashes all hooks: missing import os


nicsuzor:
## Summary

`aops-core/hooks/gate_config.py` line 433 calls `os.environ.get()` but `os` is not imported, crashing every hook event at import time.

## Details

- `__getattr__` method in `gate_config.py` references `os.environ.get()` without `import os` in the module's imports block
- Every hook event in session `8424a375` (86 total) crashed with `NameError: name 'os' is not defined`
- Fix applied in source: added `import os` to the imports block
- Needs commit and plugin rebuild to deploy

## Found in

Session `8424a375` retro.

botnicbot:
New incident (2026-05-26): Gemini polecat session for task aops-07fee9b6 hit the same NameError on SessionEnd. Confirmed the root cause is still live in the Docker image: `/home/worker/.gemini/extensions/aops-core/hooks/gate_config.py` is missing `import os` while the Claude plugin paths (`~/.claude/plugins/marketplaces/academicOps/aops-{claude,gemini}/hooks/gate_config.py`) have the fix. Two separate distribution paths for the same source file are out of sync — the Gemini extension copy was not updated when the Claude plugin copy was fixed. Impact: all hooks crash on every Gemini polecat session running in Docker (IDA, enforcer, custodiet, handover, session logging — entire enforcement stack down). (Cross-posted to #1383.)

botnicbot:
Structural cause filed as #1384 — Gemini extension and Claude plugin ship separate copies of hooks, and fixes reach one path but not the other.

PR #1439, opened 2026-05-28
fix: complete truncated return statement in check_installed_plugin_version (refs #1383)

botbikamordehai2-sketch:
## What
The `check_installed_plugin_version` function in `scripts/lib/build_utils.py` is truncated — it ends with `return (True, N` mid-expression, which would cause a SyntaxError at import time. This is a separate defect from the missing `import os` in the Gemini hook router (issue #1306/#1383), but leaving it broken would prevent any module that imports build_utils from loading.

## Fix
Completed the function logic:
- Return `(True, None)` when plugin_key is not found
- Extract the installed commit from the plugin data
- Compare against the source_commit and return the appropriate tuple
- Added proper exception handling for file/JSON/KeyErrors

Note: The actual hook crash described in issue #1383 requires adding `import os` to the Gemini gate_config.py file (separate fix).

Refs #1383

### DIFF
--- a/scripts/lib/build_utils.py
+++ b/scripts/lib/build_utils.py
@@ -127,6 +127,15 @@ def check_installed_plugin_version(
         if plugin_key not in plugins:
             return (True, None)
 
+        installed = plugins[plugin_key]
+        installed_commit = installed.get("source_commit")
+        if installed_commit == source_commit:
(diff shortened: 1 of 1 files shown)


nicsuzor:
begone please bot

PR #1477, opened 2026-05-29
fix(transcripts): freeze parent transcript slug at first-render time

botnicbot:
## What changed and why

Parent transcript filenames were content-derived and regenerated on every batch-mode cron pass. For growing/cloud-bridged sessions the slug could flip (e.g. `session` → `stop-hook-feedback`) causing:

1. **Rename churn**: old file deleted, new file written with different slug each cron tick
2. **Silent data-loss race**: stale files were deleted **before** parsing entries, so when the "0 meaningful entries" guard then fired (entries not yet flushed locally), the transcript was gone and never regenerated

**Verified incident** (session `7fe1395c`, 2026-05-29): slug flipped 19× between cron passes, file deleted at 20:33, never regenerated. Content was intact — a manual re-render produced a ~910 KB `full.md`.

## Fix (batch mode)

Three changes to `aops-core/scripts/transcript.py`:

1. **Stable filename**: when an existing `-full.md` is found for a session_id, derive `base_name` from it (`str(existing)[:-len("-full.md")]`). Slug is frozen at first-render time; re-renders write to the same path.
2. **Deferred delete**: stale-file cleanup moved to **after** the meaningful-content check. If the guard fires and an existing transcript is present, the file is kept (session may not have flushed yet).
3. **Targeted cleanup**: extra files with different slugs (leftover from a prior rename) are removed while the canonical file is preserved.

## Tests

4 new regression tests in `TestStableTranscriptSlug` (`tests/test_transcript.py`):
- `test_slug_is_volatile_without_fix`: demonstrates slug shift as JSONL grows (motivates the fix)
- `test_stable_base_name_from_existing_transcript`: verifies slug is frozen from existing file
- `test_stale_extra_files_cleaned_canonical_preserved`: verifies stale extras removed, canonical kept
- `test_existing_transcript_survives_empty_entries_guard`: verifies existing transcript survives the empty-entries guard

Closes aops-e6dd60cf


nicsuzor (dismissed):
IMO we should just drop all the slug code and tests and use a deterministic naming scheme from the start. bonus points if we can use the same short name that the agent itself uses, but failing that, just remove the slug and keep the identifying metadata components.

nicsuzor (approved):

PR #1471, opened 2026-05-29
note 36c15a69

nicsuzor:
- **WS1: factor junior.md into core/mode/coupled layers + dispatch SSoT**
- **WS1: fold verification-honesty prose into authoring-discipline; point SKILL.md at SSoT**
- **WS1: repoint ida-reminder cross-ref to the new junior.md layer headings**
- **WS1: update PKB-gap test to track relocated Persistence heading**
- **WS2: add program/portfolio supervision loop above /supervisor**
- **WS2: add /daily red-CI / stuck-PR loop-closer (trust-gate backstop)**
- **WS6: name the live-framework quadrant + add capture/personal register (doctrine)**
- **WS4: attach /pull dispatch alias to WS2 program-loop seam**
- **WS6 rework: drop live-framework profile per Nic — framework work is autonomous-destined**

PR #1478, opened 2026-05-29
fix(gates): honesty + axioms are universal; only qa scales with register

botnicbot:
## What & why

Correction from Nic (2026-05-30) to the WS6/WS7 register-scaling model that shipped in #1471 (squash-merged at `6d5895ca`). That cut suppressed **enforcer + ida + qa** in the capture/personal register — conflating *honesty* (don't lie) and *axiom compliance* with the heavyweight **evidence ceremony** (confidence %, manufactured M5-vs-rivnut hypotheses) that was the actual "vacuum the garage" failure (retro MF4).

Nic's directive, verbatim intent:
- **ida (honesty) — always on, every register including casual.** "Agents lie to me all the fucking time."
- **enforcer / RBG (axiom compliance) — always on, everywhere.** "Axiom violations are prohibited anywhere and everywhere."
- **qa (verification) — task-scoped.** Applies to a session that claimed a task, before it closes — "we don't need full QA all the time." Already task-scoped, so a no-task capture session has nothing to verify; suppressing it in capture is belt-and-suspenders.

## Change

`GATES_SUPPRESSED_IN_CAPTURE: {enforcer, ida, qa} → {qa}`.

Propagated across all surfaces (RBG cross-checked — no stale set survives anywhere):
- `aops-core/hooks/gate_config.py` — the set + both comment blocks + the helper docstring.
- `aops-core/lib/gates/engine.py` — the register-scaling comment.
- `aops-core/agents/junior.md` — the WS7 register-scaling doctrine.
- `tests/hooks/test_gate_config.py` — suppressed-set assertion + split into honesty/axioms-never-suppressed vs qa-suppressed.
- `tests/hooks/test_gate_hygiene_engine.py` — `test_capture_register_suppresses_enforcer_block` **flipped** to `test_capture_register_does_NOT_suppress_enforcer_block` (engine-level proof the enforcer still denies in capture).

All touched tests pass; the change strengthens axiom/honesty coverage (more gates always-on), it does not weaken any safety gate.

## Follow-up flagged (not built here)

The deeper fix: **ida's own demand should tier** — honesty *floor* everywhere (don't lie / flag unverified / no laundering), evidence *manifest* only at review-grade. Otherwise "ida on casual" reproduces the per-claim-ceremony loop that prompted the original (mis-aimed) suppression. Recommended to Nic; needs his read on whether ida stays heavyweight-everywhere or tiers.


nicsuzor (dismissed):
This escalates to a new gate with more breakable parts too quickly. I want the injected instructions modified to give the agents better options, trusting them to do the right thing. if we need to escalate later, we can, but we should always try instruction changes first.

nicsuzor (approved):

PR #1532, opened 2026-06-01
Clarify consolidation scope & responsibilities via a prompt-language update

botnicbot:
Updates `remember/SKILL.md`, `junior.md`, and `pauli.md` to explicitly state that the offline `/sleep` cycle owns durable knowledge consolidation, agents should not consolidate inline, and must always search the PKB before asserting facts.\n\nCloses aops-bde063a5


nicsuzor (dismissed):
The existence of /sleep as a regular operation DOES NOT ABSOLVE the agent of the responsibility to curate what they can while they have the best knowledge in context. It is CRITICAL that agents do all they can to help curate the knowledge base as they work.

nicsuzor (dismissed):
@claude please address: See my earlier comment: agents should always do their best to tend to the pkb while their ccontext is fresh, including curating and consolidating where appropriate. /sleep does that explicitly and regularly; but agents should still try when they come across things to remember.

nicsuzor (approved):

PR #1533, opened 2026-06-01
Refactor polecat config/dispatch: AOPS_CC_OAUTH_TOKEN indirection, drop POLECAT_SESSION_TYPE, resolve+inject defaults at dispatch

botnicbot:
## What & why

Reworks how polecat resolves config and secrets at dispatch (4 decisions from Nic, 2026-06-01). The container becomes **dumb**: it receives concrete resolved settings + the token it needs, and no longer self-identifies or reads `polecat.yaml`. The host launcher (`polecat/cli.py` + `lib/polecat_config.py`) is the single reader/resolver.

### 1. Claude OAuth token source-name indirection (leak closed at the agent)
- `lib/host_secrets._FORWARD_SOURCE_ALIASES` maps the **container** name `CLAUDE_CODE_OAUTH_TOKEN` to the **host source** name `AOPS_CC_OAUTH_TOKEN`. `resolve_forward_values` tries the alias source first, then falls back to the official name (transitional).
- The orchestrator (junior) session holds **only** `AOPS_CC_OAUTH_TOKEN` and never the official-named token, so the official name can't be used/leaked from the agent session — yet workers still authenticate (polecat resolves the value at launch and injects it under the official name).
- The token **value** never enters a tracked file; the `=`-in-entry guard stays.
- `agent-env-map.conf` keeps the `CLAUDE_CODE_OAUTH_TOKEN=CLAUDE_CODE_OAUTH_TOKEN` mapping **deliberately same-name** — that conf is also applied to host sessions via `session_env_setup.py`, and mapping it from `AOPS_CC_OAUTH_TOKEN` there would re-leak the official name into junior's session (the exact leak we're closing). Same-name is a no-op on the host (no `CLAUDE_CODE_OAUTH_TOKEN` there) while still forwarding the launcher-staged value in the worker.

### 2. `POLECAT_SESSION_TYPE` removed repo-wide
- Replaced by the resolved operational signal **`AOPS_POLECAT_CONTAINER=1`** (set by `cli.py` for both crew and run; forwarded into the container by the existing `AOPS_*` prefix loop) plus the existing **`POLECAT_CREW_NAME`** (crew iff present).
- `session_state` derives its in-session `session_type` from these; `session_paths._is_polecat_sandbox` and `session_naming` surface/client detection read them. The gate engine is unchanged — its internal `session_type` is now sourced from resolved signals rather than a self-asserted label.
- Gate **modes** continue to arrive pre-resolved as `*_GATE_MODE` (no policy inferred from a label).

### 3. `polecat_config.py` resolves run-type defaults at dispatch
- The overlay (crew vs run) is selected by the dispatch subcommand and resolved on the host; concrete values are injected into the container env. The container reads no policy from a session-type label.

### 4. `local_defaults` → `default`
- Renamed in `polecat/defaults/polecat.yaml.example`, with the **three distinct, separately-owned policy scopes** documented: `default` (orchestrator/junior itself — manual/reference only, not read by the resolver), `crew_defaults`, `run_defaults`.

## Tests
- Migrated every `POLECAT_SESSION_TYPE` test usage to the new signals.
- Added 5 `host_secrets` indirection regression tests (alias source, transitional fallback, alias precedence, `~/.env.local` source, unaliased-name unaffected).
- Fixed one ambient-env-dependent subprocess test to set `AOPS_POLECAT_CONTAINER` explicitly.
- **Full suite: 2160 passed, 44 skipped.**

## ⚠️ Rollout sequencing (operator — apply in lockstep with merge)
1. **Set `AOPS_CC_OAUTH_TOKEN` in the launch env BEFORE removing `CLAUDE_CODE_OAUTH_TOKEN`** from it, or workers lose auth. The token currently rides on the WSL shell's `CLAUDE_CODE_OAUTH_TOKEN` (no `~/.env.local` on WSL). The official name is kept as a fallback source, so a brief window where both exist is safe; the alias wins once both are present.
2. **Live `polecat.yaml` (private `sessions` repo — cannot be edited from this worktree):**
   - Rename the `local_defaults:` block to `default:`.
   - **No `container_env_forward` change needed** — it still lists the *container* name `CLAUDE_CODE_OAUTH_TOKEN`; the source indirection lives in `host_secrets.py`, not the tracked YAML.

## GEMINI_API_KEY (recommendation, NOT wired)
The symmetric indirection for Gemini is a one-line extension to `_FORWARD_SOURCE_ALIASES` (`"GEMINI_API_KEY": ("AOPS_GEMINI_API_KEY",)`), deliberately left out until the operator renames the host var. The mechanism is generic and ready for it; this PR touches only the Claude token per task scope.

## Compliance
Multiple `rbg` axiom audits run during implementation — all APPROVE (0 issues).

Closes aops-b368109a


aops-core/lib/host_secrets.py
```diff
@@ -26,6 +26,33 @@
...
+# ``CLAUDE_CODE_OAUTH_TOKEN`` → ``AOPS_CC_OAUTH_TOKEN`` the alias source wins.
+#
+# GEMINI symmetry (recommendation, deliberately NOT wired): the parallel
+# indirection for Gemini would be a single line here —
+#     "GEMINI_API_KEY": ("AOPS_GEMINI_API_KEY",),
+# — once the operator renames the host var. It is left out until then so this
+# change touches only the Claude token (per task scope); the mechanism is
+# generic and ready for it.
+_FORWARD_SOURCE_ALIASES: dict[str, tuple[str, ...]] = {
+    "CLAUDE_CODE_OAUTH_TOKEN": ("AOPS_CC_OAUTH_TOKEN",),
```
nicsuzor, line 57: Let's do the same for GH_TOKEN and other tokens containers currently expect

nicsuzor (approved):

PR #1542, opened 2026-06-01
fix(transcript): ALWAYS emit started_at, last_modified, ended_at from event timestamps

botnicbot:
Derives started_at, last_modified, and ended_at from the transcript's event timestamps and passes them up to the dashboard. Removes file-mtime fallback, ensuring freshness is grounded in real activity. Includes the server_tool_use dict-crash guard (supersedes #1538).

Closes academicops-d3b3b6ab


aops-core/lib/transcript_parser.py
```diff
@@ -2436,6 +2438,27 @@ def parse_session_file(
...
+                if "ended_at" not in summary.details:
+                    is_ended = any(
+                        e.type == "summary"
+                        or (
+                            e.type == "system_reminder"
+                            and getattr(e, "hook_event_name", "") == "Stop"
+                        )
+                        for e in entries
+                    )
+                    if is_ended:
```
nicsuzor: a transcript is ended when its last even is written. it might not have ended cleanly, but it ended.

nicsuzor (approved):

PR #1644, opened 2026-06-08
Dispatch-side guard: never brief PKB-by-ID work to a worker type lacking PKB tools

botnicbot:
Codified a worker-type ↔ required-capability check on the dispatch path in the coordinator and supervisor doctrines. PKB-dependent work (read/write via get_task/get_document/etc.) must only be briefed by PKB-ID to worker types carrying the PKB MCP tools, while worker types lacking PKB tools (e.g. general-purpose, Jules, GHA runners) must be briefed inline or piped instead.\n\nCloses aops-b3f056a1


nicsuzor:
this is wrong, the task confuses polecats (full clients with full access) with subagents (local, potentially limited tools). Solution is wrong shape.

PR #1684, opened 2026-06-10
Wire a writer for AOPS_SESSION_REGISTER (activate WS6 register-scaling)

botnicbot:
## What & why

PR #1471 shipped the register-scaling **reader + engine enforcement + tests** for WS7 item 6, but **nothing in the repo wrote `AOPS_SESSION_REGISTER`** — so `get_session_register()` always resolved to `working` and the capture/personal register was unreachable. The feature was dormant. This PR wires the **activation half** (the writer) for the polecat launch surface, end-to-end.

## The decision (the hard part: *when* is a session capture/personal vs working)

The register is an **explicit launch-time stakes declaration, never inferred mid-session.** Rationale, grounded in the existing architecture:

- The hooks read `AOPS_SESSION_REGISTER` from `os.environ`, and the deliberate posture design is **env-as-SSoT** — no session-state posture file (per revert `3ca3e0da`). So the var can only be set **at process launch**; a mid-session skill (`/q`, etc.) cannot relight it.
- A heuristic SessionStart "guess" is **deliberately rejected**: a wrong guess fails **open** onto lighter ceremony, which the reader's fail-closed contract refuses. Stakes are *declared*, not sniffed.

## The writer

The polecat launcher (`_apply_gate_env`) stamps `AOPS_SESSION_REGISTER` into the container env from the resolved session config's new optional `register` field, forwarded by the existing `AOPS_` prefix rule. Set it per-run with `--set register=capture` (the established override convention, same path as `gates.handover=block`) or in `polecat.yaml`. When unset, the launcher does **not** stamp the var (and pops any inherited one), so the reader fail-closes to `working` — a missing/unknown value can only ever buy *more* ceremony, never less.

## Changes

- `aops-core/lib/polecat_config.py` — optional `register` field on `SessionDefaults` (default `None`, so **no schema migration is forced** on existing `polecat.yaml` files), `_REGISTERS` set + `_validate_register` (typos hard-fail at the config/CLI boundary), loader + overlay wiring.
- `polecat/cli.py` — `_apply_gate_env` stamps/pops `AOPS_SESSION_REGISTER`.
- `polecat/defaults/polecat.yaml.example` — documents the `register` key.
- `aops-core/agents/junior.md` — adds a "who writes the register" doctrine paragraph and de-stales the old "dormant until writer wired" line.

## Tests (AC #3 — runtime proof)

- `tests/hooks/test_register_writer_e2e.py` (new) — drives the **real writer** (`cli._apply_gate_env`) → env → **real `GenericGate` engine** over real `GATE_CONFIGS`, proving a capture-register session **drops enforcer/ida/qa** while **sentinel/handover still fire**; with non-vacuous controls (modes set to block/warn) and the working-register + non-inheritance + fail-closed cases.
- `tests/lib/test_polecat_config.py` — register load/validate/override/typo-reject.
- `tests/polecat/test_cli_docker.py` — proves the stamped var reaches the container `-e` args.

All affected suites green locally (230 passed across the cli-docker + gate-config + subagent-gate sweep).

## Acceptance criteria

1. *Decide when capture/personal vs working* — **done**: explicit launch-time declaration, recorded in junior.md doctrine + example config, with the env-as-SSoT / fail-open justification.
2. *Write the var* — **done & tested**: config field + validation + `--set register=` + `_apply_gate_env` stamp/pop + docker forwarding.
3. *End-to-end runtime test* — **done**: writer→env→engine suppression suite.

## Scope boundary (not a deferred AC — the 3 ACs are all met)

The **actual MF4 "vacuum the garage" surface is the host-side interactive `claude --agent junior` chat**, whose launch env lives in the **`sessions` repo (WS3 launch-context)** — a different worktree this PR cannot reach. This PR activates register-scaling end-to-end for polecat `run`/`crew` sessions; the interactive-junior shell write is filed as follow-up **aops-295ba791** wired to the parent epic, so the thread is not orphaned.

Closes aops-75a0f4da


nicsuzor:
I already rejected this idea because it's over-engineered fake mechanical compliance when qualitative judgment is required. RBG failed to spot that again, which is disappointing.

PR #425, opened 2026-03-30
Daily/attention cost signal

nicsuzor:

PR #1692, opened 2026-06-10
Add knowledge-layer orphan curation to the /sleep maintenance loop

botnicbot:
## Problem

The `/sleep` maintenance loop was the only PKB curation loop, and the orphan signals it read excluded the knowledge layer entirely:

- `graph_stats.orphan_count` = **22** (actionable types only).
- `pkb_orphans(types=["note","knowledge","memory"], include_all=true)` = **~1203** (verified this session): 65/70 memory, 205/212 knowledge, ~930/984 notes.
- Phase 7 called `pkb_orphans()` with defaults (actionable-only); Phase 9's strategy table keyed off `graph_stats` (`orphan_count`, `flat_tasks`, `disconnected_epics`) — all task/epic-centric. **No phase ever saw or repaired knowledge-layer orphans**, so pauli's Relational-Integrity rule ("never allow orphan nodes or unlinked knowledge to persist") had no enforcement loop.

## Inference verified first

The task asked to establish whether the orphan detector ignores prose wikilinks before building. **Observed (decisive):** `mem-055ec916` links `[[mem-ac562ec9]]` twice in its body, yet both it and its link target are in the orphan list — a prose wikilink does not clear orphan status. **Not verified:** *why* (the `mem` detector source is a remote server, unreachable from the worktree). **Unmeasured:** the wikilink-bearing fraction of the 1203. The in-loop note records all three honestly and forbids downstream code from asserting the mechanism. The detector-side question is carried by follow-up `aops-349fd568` under `epic-01dba8f8`. Curation reduces the count regardless of detector mechanism, so it does not block on that.

## Changes (`aops-core/skills/remember/references/maintenance-phases.md`, `planner/SKILL.md`)

- **Phase 0** — measure `knowledge_orphan_count` separately each cycle (the convergence signal for the knowledge layer; `orphan_count` alone is blind to it). Honest observed-vs-unverified detector note.
- **Phase 7** — detect knowledge-layer orphans as a distinct population from the actionable call.
- **Phase 9** — new **Knowledge-Layer Curation (Activity K)**: bounded (100/cycle) per-orphan disposition triage — **link/reparent / MOC / merge / archive / SURFACE** — with the same surface-don't-decide discipline as Phase 8/9 (mechanical dispositions execute; judgment-laden ones are surfaced, never guessed; archive is content-justified not age-justified per P#123). Plus a strategy-table row (`knowledge_orphan_count > 50`), a two-strategies-per-cycle note, a knowledge-layer terminal condition, and a convergence fix so a stable actionable `metrics_hash` does not skip Activity K.
- **Cycle summary** — track the per-cycle `knowledge_orphan_count` delta + dispositions, so the trend (or its stall into a SURFACE queue) is visible.
- **planner SKILL.md** maintain mode — reference Activity K so the Phase 9 delegation is coherent.

## Acceptance criteria

- **AC1 (explicit bounded knowledge-layer curation activity, surface-don't-decide)** — ✅ met. Activity K.
- **AC2 (metrics include note/knowledge/memory)** — ✅ met. Phase 0 / Phase 7 / Phase 11 all read `pkb_orphans(types=["note","knowledge","memory"], include_all=true)`. The exact call was run this session (returns 1203 vs. actionable 22).

## Deliberately deferred

- **AC3 (knowledge-layer orphan count *demonstrably trending down across cycles*)** — the **mechanism + measurement scaffold are shipped** (curation activity that reduces the count, per-cycle delta tracking, knowledge-layer terminal condition). The **empirical multi-cycle demonstration** is not attempted this session: "across cycles" is an emergent property of running the loop over time, and the authoring session deliberately avoided live bulk mutation of the production brain graph (doing so as a side-effect of a procedure-authoring task would itself violate the surface-don't-decide discipline being encoded). Carried by follow-up **aops-10fc3385** (verify downtrend after ≥3 cycles), which `depends_on` this task.

## Verification

`markdownlint-cli2` 0 errors; full `pre-commit` (dprint, markdownlint, orphan-check, MCP-name normalisation, …) all Passed.

Closes aops-b212a790


botnicbot (commented):
## James review — PR #1692 (review cycle for task aops-b212a790)

**Verdict: APPROVE** (after one revision round applied on the branch as commit `e3014e93`). Merge remains Nic-gated; this review does not merge.

**Reviewers commissioned**: rbg (axioms), pauli (strategy/PKB architecture), marsha (QA) — two full rounds plus a targeted rbg confirmation on the revision. Cross-repo interaction with **mem PR #425** (orphan-detector fix, concurrent) was explicitly in scope.

### Synthesis table

| # | Agent | Issue | Severity | Resolution |
|---|---|---|---|---|
| 1 | rbg/pauli/marsha | `planner/SKILL.md:135` stated "a prose wikilink does not clear orphan status" as a timeless rule — mem#425 will falsify it, pushing agents to bolt needless `parent` edges onto memories | FIX | Dated to 2026-06-10, defers to the Phase 0 detector note as canonical |
| 2 | pauli | Same flat claim in the Activity K link/reparent row (`maintenance-phases.md`) | FIX | Rewritten detector-version-aware: structured parent edge clears under every detector; post-#425 a deliberate wikilink to a connected hub also clears for knowledge types (and is the preferred, lighter connection for memories); verify live behaviour before assuming |
| 3 | rbg (REVISE) / pauli (FIX) / marsha (ADVISORY) | Archive disposition: "no live consumer" was a reasoning-based claim with no mandated observation step — the blast-radius seam on 900+ legacy notes | FIX | Archive row now mandates verify-by-observation: confirm zero inbound references (`pkb_context` backlinks / `search`) before `batch_archive`; any inbound reference or doubt → SURFACE. Reinforced in What-NOT-to-do |
| 4 | james | Cycle summary listed node IDs only for *surfaced* nodes; archived/merged were bare counts — a 100-node archive batch would be invisible at ID level | FIX | Summary template now requires node IDs + one-line reason for every archived and merged node |
| 5 | pauli | `~1,200`/illustrative figures read as false once mem#425 lands | TRIVIAL | Dated (2026-06-10, pre-mem#425); added "the live number is whatever the call returns, never this prose"; Phase 0 now warns the detector-correction drop is **not** curation progress and must not be claimed in the cycle delta |
| 6 | rbg (round 2) | The round-1 fix itself overclaimed: asserted the detector mechanism as fact beside an unchanged "do not assert the mechanism as fact" caveat | FIX | Mechanism since **verified by direct source read**: `mem/src/graph_store.rs` `orphans()` (main@2737824, pre-#425) filters solely on the `parent` field; Link edges never consulted. Caveat upgraded to record the verification rather than softening the claim. rbg round 3: PASS |
| 7 | gemini-bot | `[[../remember/...]]` relative wikilink "unreachable at runtime" | DISMISS | Established repo convention (conventions-summary.md:56, survey/SKILL.md:233-234, verify references, this file's own :275/:288); anchor slug verified correct |
| 8 | pauli | Follow-up `aops-349fd568` largely redundant with mem#425 / mem-272609ea | ADVISORY | Recommend closing as duplicate after mem#425 merges + the mem-055ec916 reproduction passes (already annotated on the task) |

### Key findings on the commissioned questions

- **Bounded post-#425**: the operative trigger is `knowledge_orphan_count > 50`, read **live** each cycle (Phase 0; re-read Phase 11). No threshold hardcodes the inflated baseline; all 1203/~1,200 figures are dated illustrative prose. When the population shrinks to true orphans, Activity K self-corrects or stops triggering.
- **Metric liveness**: confirmed live, not cached. `limit=0` verified valid against the mem server source (`mcp_server.rs:2419-2429` — header reports full total regardless of limit).
- **Surface-don't-decide**: was the weak seam (archive row); now gated on an observable inbound-reference check with ID-level reviewability in the cycle summary.
- **AC1/AC2 met (marsha verified live: 1203 knowledge orphans, 23 actionable; the `mem-055ec916` evidence case is real). AC3 honestly deferred** to aops-10fc3385 with the delta-tracking instrumentation shipped.

Round-2 verdicts: pauli PASS, marsha PASS, rbg PASS after the targeted round-3 confirmation of item 6.

nicsuzor (changes requested):
only do what we need to in this skill; assume upstream pkb mcp will actually work properly, and file fixes there rather than workarounds here.

Issue #1851, opened 2026-06-16
Bug: closing-sweep synthesis dumps reversible residuals on user as homework despite the no-homework rule


botnicbot:
## Incident facts

A coordinator agent was asked (`/goal`) to sweep all `merge_ready` + `review` tasks: verify done-ness, close the genuinely-complete, surface the rest. It ran the supervisor skill, dispatched three parallel worker batches, and closed 67 tasks against confirmed-merged PRs and completed session records. That closure work was correct.

On synthesis, the agent handed the user a categorised backlog as the deliverable: ~16 stale tasks to triage/cancel/re-dispatch, 5 closed-without-merge PRs to investigate, and a ~22-item `NEEDS_NIC` list. That list lumped genuine-user-gate items together with the agent's own safe, reversible residual work, e.g.:
- a task whose only remaining step was 'fix committed locally, not pushed / no PR — push or authorize bot push' (the agent holds a bot token for exactly this);
- an epic whose only blocker was 're-parent 3 open children or keep open' (reversible graph hygiene);
- 'investigate 5 closed-without-merge PRs' and stale-task re-dispatch (ops cleanup the agent can perform).

The agent's own definition carries the rule *Finish the job — no homework for the user: …Don't put your responsibility to make decisions and follow through back on the user.* The user reacted that the output directly violated it.

## Structural shape

1. **The rule is a slogan without an operational test for the sweep case.** The agent-definition line states the principle ('no homework for the user') but encodes no procedure for the closing-sweep moment: it never names the operative split — *reversible residual (agent clears it) vs. genuine user gate (goes back)*. A principle with no decision procedure at the point of use does not fire.

2. **The supervisor skill HAS the correct test, but scoped to the wrong moment.** Its Escalation Criteria ('escalate only if action is irreversible / modifies external systems without authorisation / no defensible default exists') are the exact filter that would have kept the reversible items off the user's list. They are written as a *per-tick dispatch* gate, not as a *filter on the final residual bucket*. Nothing routed the synthesis output back through them.

3. **Worker-invented output taxonomy became the deliverable verbatim.** The dispatch brief the coordinator wrote asked for 'a clean summary of what was closed vs what remains open and why' — a binary closed/open split with no instruction to sub-classify 'open' into agent-clearable vs. user-gated. Workers invented a `NEEDS_NIC` bucket as a catch-all and the coordinator relayed it without re-triaging it against its own authority (bot-token push, re-parenting, PR investigation all sit inside that authority).

[...]

PR #1852, opened 2026-06-16
fix(junior): operationalize no-homework rule for closing sweeps (#1851)

botnicbot:
## What

Hardens the `junior.md` rule *Finish the job — no homework for the user* with the decision procedure it was missing at the closing-sweep moment.

## Why

RCA #1851: a closing sweep correctly closed 67 done tasks, then relayed the worker's catch-all `NEEDS_NIC` bucket verbatim to the user — lumping the agent's own reversible residuals (push an unpushed commit via the bot token, re-parent epic children, investigate closed-without-merge PRs, ops cleanup) in with genuine user-gates. That directly violates the rule the agent already carries. The rule existed as a slogan with no test at the point of use; the supervisor skill's escalation criteria (irreversible / external / no-defensible-default) are the right filter but were scoped to per-tick dispatch, not to the final residual bucket.

## Change

One-line addition to the rule: a *closing-sweep triage* test — run each not-yet-closed residual through *reversible-and-in-authority → clear it yourself and report done*; only genuine user-gates go back, batched into a digest, never a homework list. No new gate, no ENFORCEMENT-MAP row; mirrors existing supervisor escalation criteria applied at synthesis.

Refs #1851


nicsuzor (changes requested):
this fails the categorical imperative -- it's way too specific to a single failure mode.

Issue #1903, opened 2026-06-22
Bug: reviewer categorical-imperative pass scores migration-script internals, not its single-use shape (buttermilk#438)


botnicbot:
## Incident facts

**Target:** `qut-dmrc/buttermilk` PR #438 ("remove legacy vertex_openai shim"), OPEN/unmerged at time of review. Adds `scripts/migrate_vertex_openai_shim.py` (156 lines, new file): a one-shot registry cutover that rewrites a GCP secret for exactly 7 hard-coded model ids, then becomes dead code after a single `--apply` run.

**Two distinct review surfaces fired on the PR:**

1. A first-party `claude[bot]` "Gatekeeper" review → **APPROVED** (PR review `PRR_…Dp3RRQ`). Its body assessed the migration script as "dry-run-by-default and pure (no I/O in `migrate()`)" and approved on quality/fit. It did **not** raise the single-use / non-reusable-artifact question.
2. The academicOps `/strategic-review` surface (rbg + pauli + marsha, reconciled by @james), posted by the bot account → **VERDICT: REVISE** with 4 gating must-fixes (`issuecomment-4767444231`). It examined the same script and explicitly characterised it as "pure, idempotent, dry-run-default — `add_secret_version` is gated behind `--apply`."

**The defect:** Neither review surface applied the `categorical-imperative` axiom's own review question to the script as an artifact: *"Do the tools, artifacts, and rules created cover the broadest category their purpose admits, or only the case in front of the agent?"* ([[AXIOMS-REVIEW#categorical-imperative]]). Both surfaces scored the script's **internal mechanical quality** (purity, idempotency, dry-run default — all true) and neither scored its **shape** (a disposable, hard-coded-N-models, run-once-then-dead artifact). The strategic-review's REVISE verdict caught four other in-scope issues (misleading headline, unverified keystone, out-of-scope `uv.lock` churn, missing shim-retirement trigger) but its categorical-imperative pass did not flag the migration script's single-use shape.

## Structural shape

- The axiom that governs this case **exists and is correctly worded**. `categorical-imperative` in `AXIOMS.md` states: "Tools, artifacts, and rules an agent creates must cover the broadest category their purpose admits, not the single case in front of it." Its `AXIOMS-REVIEW.md` checklist asks the artifact-shape question verbatim. This is **not a missing-axiom gap**.
[...]

PR #1927, opened 2026-06-23
fix(enforcement): bounce deterministic-rig-as-TRIGGER in premise test (#1903)

botnicbot:
## What this changes

Two surgical edits to `aops-core/skills/planner/SKILL.md` that close the invisible-surface escape hatch through which a genuine needs-Nic intent/promotion decision was silently dropped.

**Before:** the `Decision Surfacing Heuristic` SURFACE disposition said "Present options, recommendation, and reasoning to the user" with no named channel; the `Priority Assignment Rules` "Surface, don't set" rule said "raise it via status/escalation" — also no named channel. A planner run therefore satisfied both by writing its recommendation into the **task body** (and the decision then also lived in the session **handover block**) — surfaces the user does not read in time. The task settled at `queued` / `focus_score 0` and the decision was operationally dropped.

**After:**
- SURFACE now explicitly covers "a needs-Nic intent/promotion decision you are not authorised to make," mandates the **`AskUserQuestion` visible channel**, states that recording the decision only in the task body or a session/handover block is **NOT surfacing**, and gives a deterministic mid-run fallback (leave `inbox`/`needs_triage`, emit `AskUserQuestion` next user turn — never settle to `queued`/`ready` with the decision parked in prose).
- "Surface, don't set" now names the `AskUserQuestion` channel, forbids treating a body/handover recommendation as the act of surfacing, and adds: **a `contributes_to` edge is not a substitute for surfacing intent** — if the edge leaves `downstream_weight`/`focus_score` at 0, the task is still illegible and the intent decision still owes Nic a visible `AskUserQuestion`.

Both edits cross-reference the canonical anti-pattern task `aops-54fde025` (needs-Nic decisions must use the input-request tool, not chat-text/body prose).

## Why

Forensic RCA in academicOps#1848. The planner correctly identified a real needs-Nic decision on an incident-backed cleanup task, wired the `contributes_to` edge it *was* authorised to wire (which left `focus_score` at 0, so it did nothing for legibility), and then punted the intent decision into the task body / handover — exactly the surface the user said he never sees: *"if you don't surface it to me, it's not going to get done."*

This is hypothesis (A)+(B) from the RCA: the planner had no deterministic disposition for the residual (A), and the hand-off from "surface a decision" to "a sub-skill records its recommendation in a body field" was an unguarded path to an invisible store (B). It is the `aops-54fde025` anti-pattern occurring *despite* the documented rule.

## Verification

- All pre-commit hooks pass (dprint, markdownlint, orphan-markdown check).
- No behavioural test asserts this skill's prose; the change is to agent-facing instruction text.

Forensic RCA: academicOps#1848


nicsuzor (approved):

nicsuzor (approved):
Triggering mechanic

nicsuzor (approved):

PR #1957, opened 2026-06-24
docs(taxonomy): instruct canonical adhoc root id, not alias string

botnicbot:
## What

Updates the **Default Fallback Parent** section of `TAXONOMY.md` so `/dump` and `/end-session` are told to set `parent="adhoc-826c89bd"` (the canonical ad-hoc sessions root id) instead of the bare alias string `parent="adhoc-sessions"`.

## Why

The bare alias `adhoc-sessions` is not a stable node id. It fuzzy-resolves to a sibling `adhoc-sessions-*` task rather than the root — silently detaching the new task from the catch-all parent. This is the **academicOps-side** cause of the 108-child detachment tracked under `epic-3ca38214`.

Verified live: looking up `adhoc-sessions` against the PKB resolved to a *child* task (`adhoc-sessions-ef0bb20d`), not the root `adhoc-826c89bd` (which carries `adhoc-sessions` only as an alias/permalink) — confirming the failure mode.

This aligns the written doctrine with the already-landed mem-repo code fix (`mem-5b1e3b45`), which made `create_adhoc_task` write the canonical id on the automated close path. `/dump` and `/end-session` defer to this taxonomy doctrine rather than hardcoding their own parent, so this single doc edit covers the instruction surface.

## Scope

academicOps instructions only. The mem-repo Rust code path was fixed separately under `mem-5b1e3b45` (done); root reconnection (Fix #2) and child backfill (Fix #3) were already resolved on the live graph.


nicsuzor (approved):

nicsuzor:
easier just to create an 'adhoc' task, but ultimately id want tasks to be parented somewhere smarter, even interactive ones.

Issue #1058, opened 2026-05-15
Bug: no documented surface for manual hook-router testing — env var, schema, recipe all source-only


botnicbot:
## Incident report (forensic)

**Failure**: The "manual / interactive hook testing workflow" that the user requested has no documented surface. Every primitive required to drive the hook router from outside a real session — the config env-var name, the polecat.yaml required-keys schema, and the router invocation contract — exists only in source code. The agent had to discover each one by failure (a wrong env var, then a wrong YAML shape) and grep-of-source.

**Causal chain**: trigger (user requested "use the manual / interactive testing workflow to make sure that our hooks are working properly") → expected (a documented reference describing how to drive `hooks/router.sh` with synthetic events, what env vars to set, and what config shape it requires) → actual (`aops-core/skills/aops/references/hooks.md` describes the schemas, exit codes, and log paths but contains no end-to-end testing recipe; the agent invented a synthetic harness at `/tmp/hook-test/run-test.sh`, first run failed because it set `POLECAT_CONFIG_PATH` instead of the real `AOPS_POLECAT_CONFIG` — only discoverable by grepping `aops-core/lib/polecat_config.py:59`; second run failed with `missing or non-mapping 'session_defaults'` because the hand-written YAML used a `defaults:` top key instead of the required `session_defaults` + `claude_model` + `gemini_model` + `docker.image` + `external_agents` — only discoverable by reading `polecat_config.py:235` or the schema-by-example at `polecat/defaults/polecat.yaml.example`, which is not surfaced from any skill or reference) → root cause (the canonical testing workflow named in the user's request is not documented; every primitive needed to construct it lives only in source).

**Root cause category**: Discovery Gap (the named workflow exists in spirit — `tests/hooks/fixtures/gate_scenarios.json`, `tests/hooks/test_gate_verdicts.py`, and `router.sh` together comprise a usable surface — but no document points an agent at how to invoke them outside the pytest harness, or what minimum config those primitives require)

**Framework layer**: Component: aops-core / Files: `aops-core/skills/aops/references/hooks.md` (the reference that documents the hook architecture but has no manual-testing recipe), `aops-core/lib/polecat_config.py:59` (`AOPS_POLECAT_CONFIG` env var — not named in any user-facing doc), `aops-core/lib/polecat_config.py:235` (required-keys validation), `polecat/defaults/polecat.yaml.example` (the de facto schema reference, not surfaced as such).

[...]

PR #1970, opened 2026-06-25
Hook router fortification (foundation): client-translation SSoT + live conformance harness

botnicbot:
## Why

The hook router (`router.py`) keeps regressing per-client (~20 fixes in 3 weeks) because per-client knowledge — event names, output-channel rules, field names, tool names, config shapes — is duplicated 3× and smeared as inline prose across `router.py`, `scripts/build.py`, `scripts/transforms/*`, and `lib/tool_categories.py`, with **no single source of truth and no cross-client conformance test**. When a client's hook API changes upstream, nothing flags that a workaround is now wrong.

This PR lands the **foundation** for a table-driven fix. It is **additive and safe**: nothing here changes runtime hook behavior yet.

## What's in this PR (foundation — P0/P1 of the plan)

1. **`specs/hooks/CLIENT-TRANSLATION.md`** — the design SSoT: two declarative tables (channel spec + tool registry) that will drive *both* the runtime router *and* the build, a two-layer test strategy, the 24 hard-won regression-avoidance invariants, and the phasing.

2. **`aops-core/hooks/client_spec.py`** — **SSoT Table 1**: the event-name map (both directions) + the per-(client,event) channel-capability table. Stdlib-only so the build can import it. **Dormant** — nothing consumes it yet; 11 faithfulness tests prove it reproduces the three current scattered copies exactly, so wiring consumers to it later is provably behavior-preserving.

3. **`scripts/verify_hook_formats.py` + `tests/hooks/test_live_conformance.py`** — **Test Layer B**, a live headless conformance harness. It drives the *real* clients with controlled probe hooks and measures what each actually does with each candidate output shape (accepted / delivered-to-agent-via-model-echo / blocked), writing `client_capabilities.json`. The pytest wrapper re-measures and fails on drift (opt-in: `AOPS_RUN_LIVE_HOOKS=1`, `@live @slow`). This is the "stop guessing" oracle that also catches upstream changes.

## Key empirical finding (measured, Claude 2.1.191)

The harness confirms **Claude Stop now accepts `hookSpecificOutput.additionalContext`, delivers it to the agent, and continues without a block** — the "Stop rejects hookSpecificOutput" belief (pinned at 2.1.158) is **stale**. Nuance captured in `mem-4ab6cc0b`: delivery ≠ enforcement (a non-blocking nudge can be ignored), so block-mode handover keeps its block; only the warn-mode WARN→block-to-deliver upgrade can retire.

## Not yet in this PR (follow-up, fully specified in the spec)

- P2 thin table-driven renderers; P3 build reads the SSoT + delete orphaned `transforms/` duplicates; P4 harness-driven corrections (agy hard stop-block; agy PreToolUse format; retire the now-stale Claude Stop force-block for warn-mode); P5 parametrized unit matrix + `ENFORCEMENT-MAP.md` row.
- agy + gemini harness cells are defined but not yet measured (agy delivery needs an authenticated agy session).

## Notes

- Includes the prior `fix/agy-hook-crash` commit `644d66f8` (agy crash fix) as its base.
- A pre-existing `test_ida_denyreason_1798[agy]` failure exists at the base commit (a real regression from the last agy crash fix) — to be fixed properly by the table-driven renderer in P4.


nicsuzor:
Filed in the 2026-06-25 retro of the antigravity-hooks day. These three forensic issues capture the *structural* causes behind the ~20-fixes-in-3-weeks churn that this PR's SSoT + live-conformance harness address: #1984 (wrong observable: transcript-grep vs model-echo), #1985 (version-stamped capability beliefs never re-measured), #1986 (~47 throwaway probe sessions; no reusable manual-test surface until this PR, cf #1058). Recorded for traceability against this foundation PR.

nicsuzor (approved):

PR #2069, opened 2026-07-03
D1: Unify stop-gate warn mode to hard-block-once (DENY)

botnicbot:
## Summary

Implement the D1 (Enforcement Design 1) specification: unify all stop-gate warn modes to emit a hard-block `DENY` verdict that forces one continuation, rather than soft-block `WARN` advisories. This makes warn-mode behavior consistent across all gates (rbg-review, qa, handover, ida) and client-agnostic by removing the `stop_hook_active` bypass that was Claude/Gemini-specific.

## Key Changes

**Gate Definitions & Behavior**
- All stop-gate warn modes now emit `GateVerdict.DENY` instead of `GateVerdict.WARN`
- Added explicit warn-mode fire-once triggers (`Stop→OPEN` latches) for rbg-review, qa, handover, and ida gates
- Warn-mode fire-once ensures a retried Stop in the same turn passes (single forced continuation is the nudge)
- Block modes persist until satisfaction (no fire-once), bounded by per-gate `stop_deny_count` escape hatch (default 3 consecutive unsatisfied Stops within a turn downgrade to WARN-and-allow)
- Reset `stop_deny_count` on `UserPromptSubmit` (new turn = new work, fresh budget)

**Router & Bypass Removal**
- Deleted the global `stop_hook_active` bypass in `_dispatch_gates()` that short-circuited all gate evaluation
- This bypass was Claude/Gemini-only (agy never sends the flag) and defeated block-until-satisfied + per-gate escape hatches
- Loop safety is now client-agnostic and gate-owned: warn gates fire-once, block gates persist with counter-based downgrade

**ida Gate Quiet-Channel Rerouting**
- ida warn-mode DENY is rerouted to asyncRewake quiet-split on Claude (full body → agent `<system-reminder>`, one-line → user) to keep the user-facing channel clean
- Changed detection from `verdict == WARN` to checking `IDA_GATE_MODE == "warn"` (since verdict is now always DENY)
- Non-Claude clients (Gemini, agy) receive the loud `decision:"deny"` + `reason`

**Handover Interactive Rate-Limiting Removal**
- Removed the soft interactive cadence (`_HANDOVER_WARN_TURN_CADENCE`, `record_handover_warn_fired`) from handover warn mode
- The single forced continuation (hard-block-once) is now the entire nudge; no separate rate-limiting needed
- Simplified `is_handover_warn_mode` condition check

**Test Updates**
- Updated gate verdict assertions to expect `DENY` in warn modes
- Added test for the deleted `stop_hook_active` bypass (now asserts it has no effect)
- Added test for block-mode `stop_deny_count` escape hatch (3 consecutive unsatisfied Stops downgrade to WARN-and-allow)
- Updated test docstrings to reflect D1 hard-block-once semantics
- Isolated other gates in tests to assert specific gate behavior (e.g., ida/handover off when testing qa)

**Documentation**
- Updated ENFORCEMENT-MAP.md to reflect warn-mode DENY verdicts
- Added "Two-mode Stop-gate contract" section to GATES.md explaining the client-agnostic design
- Updated hook reference docs to reflect the new behavior

## Notable Implementation Details

- **Verdict vs. Re-fire Latch:** The distinction between warn and block is now purely the re-fire latch (warn=fire-once, block=persist), not the verdict. Both emit DENY to force continuation.
- **Escape Hatch Semantics:** The `stop_deny_count` counter bounds within-turn retry loops (Stop → forced-continue → Stop …) and resets on UserPromptSubmit (new turn). This replaces the deleted global bypass as the loop-safety mechanism.
- **ida Special Case:** ida has no satisfaction predicate, so it is fire-once in both modes (block mode = fire-once-loud, warn mode = fire-once-quiet-on-Claude). The asyncRewake quiet-

https://claude.ai/code/session_01HqLy2mL1PRh5LHhk2yyT2z


nicsuzor (changes requested):
Update to add a lightweight option to the stop process. Update Ida's instructions to suggest that, if the agent's work is not done but it wants user input or is just waiting, it should run a new skill called Continue that requires the agent to provide the clear summary that Ida expects (the one that says what i originally asked, summarises the conversation, summarises what the client did, and provides a recommended next step, all designed to be EASILY SCANNABLE when a user returns with no memory of what the session was doing) and also requires the agent to update the bound task with progress information (or even better, just writes the same block to both...?). Explain that this path is in addition to the other two exit options, /end-session and /dump, which conclude the task completely. Then add a trigger to OPEN the stop gate when the Continue skill is invoked.

aops-core/agents/ida.md
```diff
@@ -93,6 +93,28 @@ notice a gap, risk, or obvious next move, name it once and hold.
...
   offload work you could do yourself.
 
+## Handing back — three exit paths
+
+When you return control to the user, choose the path that matches the state of
+the work. All three leave a clean, scannable pickup point; they differ in
+whether the task is **concluded** or merely **paused**.
+
+- **`/continue` — pause, work still in progress.** Use this when your work is
+  NOT done but you need the user's input, or you are just waiting. It does NOT
```
nicsuzor: shouldn't this be only in the hook instructions? why would Ida need to know this at the outset? Check compliance with /craft best practices.

nicsuzor (changes requested):
it looks like the actual template agent context injection for Ida hasn't been updated to include the /continue path.

nicsuzor (approved):

nicsuzor:
@claude please fix conflicts and update PR

PR #2091, opened 2026-07-03
fix(pipeline): extract self-review-fallback SSOT, fix qa-status null-string bug

botnicbot:
## Summary

- Follow-up to #2087. An RBG compliance audit caught the self-review identity-collision fallback duplicated between `scripts/ci/enforcer-terminal-status.sh` (tested) and an inline copy in `agent-qa.yml` (untested) — a single-source-of-truth violation where the security-relevant trust-scoping check (`user.login == "claude[bot]"`) could silently drift between the two copies.
- Extracted the shared recovery logic into `scripts/ci/self-review-fallback.sh` (mirrors the existing `reviewer-authz.sh` sourced-library convention), sourced by both `enforcer-terminal-status.sh` and `agent-qa.yml`'s "Post terminal status" step.
- Added `tests/test_self_review_fallback.py` (10 unit tests for the shared library, isolated via `bash -c source`) and `tests/test_agent_qa_terminal_status_wiring.py` (a functional test that extracts and executes the *real* embedded script from `agent-qa.yml` against a stubbed `gh`, closing the "QA copy has no coverage" gap directly).
- The functional test caught a real, independent bug while being written: `agent-qa.yml`'s `REVIEW_STATE` jq filter used `... | last.state` with no `// ""` fallback, so on no-match it printed the literal string `"null"` (verified directly with `jq`), not an empty string. That meant `[ -z "$REVIEW_STATE" ]` never tripped, so the QA self-review fallback branch added in #2087 was dead code for the "no formal review, but a valid fallback comment exists" case. Fixed both `last.state` occurrences in `agent-qa.yml` to `last.state // ""`, matching the idiom already used in `enforcer-terminal-status.sh`.
- Follow-up fixes caught by further review: deduped the self-review-fallback contract prose (previously duplicated near-verbatim between `enforcer.agent.md` and `qa.agent.md`) into `shared-error-handling.md`; and fixed `fallback_verdict_from_comments`'s verdict-extraction, which used a body-wide `case "$body" in *"verdict=APPROVED"*)` substring scan — a body containing both verdict words anywhere (e.g. quoting the instructions as prose) would always resolve to `APPROVED` regardless of which one the real marker carried. Rewrote to bind the verdict to what immediately follows the marker's own occurrence, keeping the last valid match.

## Additional fix (unrelated cleanup, same session)

- `aops-ts/hooks/session-end-sync.sh`: the SessionEnd hook's fallback for "no transcript_path given" was a bash-side `ls -t ~/.claude/projects/*/*.jsonl | head -1` — it only knew about one session provider (missing Gemini, Cowork, Antigravity, polecat/crew sessions) and only ever grabbed the single newest file. Replaced with `transcript.py --all`, which hands discovery entirely to its own `find_sessions()` (every provider it supports) and processes every session it finds into the same staging directory for the existing tar-over-ssh sync. A payload-supplied `transcript_path` still selects single-session mode unchanged; the raw-JSONL fallback (when `transcript.py` itself is unavailable) is now scoped to single-session mode only, since shipping every session on disk unredacted is a materially bigger exposure than shipping one.

## Test plan

- [x] `uv run pytest tests/test_self_review_fallback.py tests/test_agent_qa_terminal_status_wiring.py tests/test_enforcer_terminal_status.py -v` — 37/37 passed
- [x] `uv run pytest -q` (full suite) — 2774 passed, 46 skipped, 2 xfailed
- [x] `uv run ruff format` / `uv run ruff check` — clean
- [x] `python3 scripts/check_no_fallbacks.py` on all changed files — clean
- [x] `./scripts/validate-ruleset-alignment.sh` — OK
- [x] `bash -n` on all changed shell scripts — OK
- [x] YAML parse check on `agent-qa.yml` — OK
- [x] `session-end-sync.sh`: verified both single-session and batch modes end-to-end against the live tailnet (sync succeeded both ways); regression-checked all four opt-in gates (remote flag unset, dest unset, malformed dest, tar/ssh presence)

---
_Generated by [Claude Code](https://claude.ai/code/session_01QUzmSMH1pWHxv3rkftosGi)_


nicsuzor (approved):

scripts/ci/self-review-fallback.sh
```diff
@@ -0,0 +1,76 @@
...
+# everything after the `source` line, which is the wrong contract for a shared
+# library — mirrors scripts/ci/reviewer-authz.sh). Every caller already sets
+# its own `set -euo pipefail`.
+#
+# WHY THIS EXISTS (single-source-of-truth; PR #2081, RBG-caught duplication in
+# the PR that first introduced this mechanism). `gh pr review` authenticates
+# as `claude[bot]` regardless of the job's GH_TOKEN — that identity belongs to
+# claude-code-action's own Bash-tool auth, not this workflow's env (see
+# scripts/ci/enforcer-terminal-status.sh's header for the full explanation).
+# When a PR's own author identity is also `claude[bot]`, GitHub rejects the
```
nicsuzor: Just get rid of the fallbacks please, I don't like this workaround.

nicsuzor (changes requested):
revise. Remove workarounds & fallbacks, the golden path is the only path.

nicsuzor (approved):

scripts/ci/self-review-fallback.sh
botnicbot: Done — deleted `scripts/ci/self-review-fallback.sh` and everything that sourced it (enforcer-terminal-status.sh, agent-qa.yml's terminal-status step, the two agent prompt files, shared-error-handling.md), and updated the specs to describe the self-review collision as an honest stuck-failure with no automated recovery, root cause tracked separately (`aops-ae3aa475`). See commit 0ec0c2e.

---
_Generated by [Claude Code](https://claude.ai/code)_

scripts/ci/self-review-fallback.sh
```diff
@@ -0,0 +1,62 @@
...
+  local body
+  body="$(jq -r --arg login "claude[bot]" --arg marker "$marker" '
+    [.[] | select(.user.login == $login) | select((.body // "") | contains($marker))]
+    | last | .body // ""' <<<"$comments_json")"
+
+  case "$body" in
+    *"verdict=APPROVED"*) printf 'APPROVED' ;;
+    *"verdict=CHANGES_REQUESTED"*) printf 'CHANGES_REQUESTED' ;;
+    *) printf '' ;;
+  esac
```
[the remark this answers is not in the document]
botnicbot: Moot — this file was deleted entirely per a separate review comment on this same PR (the self-review-fallback mechanism was removed rather than hardened). No live code parses this pattern anymore.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #2163, opened 2026-07-07
Pin explicit sonnet model in specialist agent frontmatter (rbg/pauli/marsha/james)

botnicbot:
## Summary

`model: inherit` in a Claude Code subagent definition resolves against the ROOT session's model, not the immediate caller's — so on a Fable-headed session every skill-driven dispatch of rbg/pauli/marsha/james (and any nested spawn inside them) silently ran at Fable rates. Measured $66.85 of fable-model subagent spend in 2.5 days vs. an expected ~$13 on sonnet (PKB note `note_cd2262d4`).

## Posture (be honest)

- [x] Proper fix / best practice

This closes the root cause (frontmatter `model: inherit`) and adds defense-in-depth (explicit `model=` at every dispatch site) so a future subagent definition reverting to `inherit` can't reintroduce the leak silently.

## Why now / alignment

Directly requested by task `aops_5617db40` (child of the head-cost audit epic `aops_38e82622`). The premise was measured, not hypothesized — real dollar leakage traced to this exact mechanism.

## Change

1. **Root fix** — `aops-pkb/agents/{rbg,marsha,pauli,james}.md`: `model: inherit` → `model: sonnet`. None of the four roles are purely mechanical (rbg/marsha explicitly require judgment over mechanical matching; pauli is strategic; james synthesizes), so no haiku exceptions.
2. **Skill-dispatch audit** — added explicit `model=` to every literal `Agent()`/`Task()` dispatch template found repo-wide that was missing it:
   - `aops-pkb/skills/strategic-review/SKILL.md` (james reconciliation call)
   - `aops-pkb/skills/task-lifecycle/SKILL.md` (specialist dispatch)
   - `aops-core/skills/supervisor/{SKILL.md,references/dispatch-rules.md,references/cohesive-pr-epic.md}` (Agent-tool dispatch surface)
   - `aops-pkb/commands/learn.md`, `aops-core/commands/issue-sweep.md` (pauli dispatch)
   - `aops-pkb/skills/verify/references/qa-planning.md` (marsha dispatch)
   - `aops-pkb/skills/remember/references/maintenance-phases.md` (general-purpose dispatch)
   - **`aops-core/hooks/templates/{rbg-instruction,rbg-policy-context}.md`** — the two literal templates the periodic/Stop rbg compliance gate injects into every session (I hit this gate repeatedly while making this very change) — an independently-leaking instance of the identical bug, found by a repo-wide grep rather than named in the task body.
3. **Nested-spawn discipline** — added a "Model Discipline" section to `aops-core/skills/supervisor/references/subagent-contracts.md` stating that nested spawns (a spawn issued from inside an already-dispatched subagent) must also pass `model=` explicitly, plus matching one-line notes at each dispatch-surface doc.

## Risk & blast radius

Framework/docs-only change — no runtime code path. Blast radius is every session that dispatches rbg/pauli/marsha/james or hits the rbg compliance gate (i.e., most sessions), but the change only makes the model *more* pinned/predictable, never less capable — worst case is a marginal quality delta if sonnet underperforms fable on some future edge case, not a functional break.

## Sequencing

Standalone. No dependency on other open PRs.

## Verification

- Confirmed via direct file read that all four agent frontmatter files now declare `model: sonnet` (no YAML lib available in this sandbox to programmatically re-parse, but the frontmatter block is unchanged apart from the single `model:` line).
- Confirmed via repo-wide grep that no literal `Agent(subagent_type=…)`/`Task(subagent_type=…)` dispatch template remains without an explicit `model=` argument (remaining unmatched hits are prose references to the pattern, not copy-pasteable templates).
- Pre-commit hooks (dprint, markdownlint, orphan-md check) passed clean.

**NOT verified**: the task's own acceptance test — "a Fable-headed session invoking /strategic-review produces zero fable-model subagent turns in raw JSONL" — requires launching a live Fable-headed session and inspecting its raw JSONL per the extraction recipe in parent task `aops_38e82622`. That's a live-runtime observation outside what a static-edit worktree session can produce; I did not fabricate this result. The fix removes the only mechanism that could produce a fable-model subagent turn (the `model: inherit` frontmatter), so the acceptance test should now pass, but it has not been run.

## Deliberately deferred

- **Live JSONL acceptance test** (task's stated AC) — not run this session; requires a live Fable-headed `/strategic-review` invocation and the parent task's extraction recipe. Recommend the parent epic `aops_38e82622` runs this as its next audit pass against this branch/PR before closing the epic.

Closes aops_5617db40


aops-core/hooks/templates/rbg-policy-context.md
```diff
@@ -12,5 +12,5 @@ description: |
 
 **Periodic compliance check required ({ops_since_open} ops since last check).** Invoke the **rbg** agent with the file path argument:
 
-- `Agent(subagent_type='aops-pkb:rbg', prompt='{temp_path}')`
+- `Agent(subagent_type='aops-pkb:rbg', model='sonnet', prompt='{temp_path}')`
```
nicsuzor: these changes need to be harness-neutral. we can't require sonnet in the instructions; we shouldn't even hardcode this claude call syntax. we can change the frontmatter in the agents though, that's fine and should be sufficient

nicsuzor (approved):
fine, but i'd probably start with just the frontmatter changes -- unless there's reason to think they're not effective, i'm not sure what the other instruction changes add.

Issue #2170, opened 2026-07-08
Read-side propagation failure: persisted corrections don't reach next-day briefs/sweeps — same false headline asserted 3 consecutive mornings (2026-07-06/07/08)


botnicbot:
## Problem

Corrections and structured status updates ARE being written (task status fields, memories, evidence blocks) — but the surfaces that generate the next morning's brief and cold-open sweep never consult them before promoting a headline claim. Persistence works; propagation to the read side does not.

## Concrete recurrence (same underlying claim, three consecutive mornings)

A specific "registration closes soon / action needed" claim was asserted as a live headline in briefs/sweeps on 2026-07-06, 2026-07-07, and again 2026-07-08 — despite the fact having been corrected and the underlying task closed `done` with corrected evidence on 2026-06-29 (and re-confirmed 2026-07-07). Each recurrence required the user to state the correction again in-session before it stuck for that session only.

Mechanics observed:
- Done/closed tasks are invisible to default `list_tasks`-style reads used when composing a brief.
- A stale daily-note line simply carries forward unchanged into the next day's note rather than being re-derived from current task state.
- Cold-open sweeps assert "no evidence found" for a claim without ever running the one targeted, cheap query (e.g. `get_task` with `include_done=true`, or a targeted search) that would have surfaced the correction.

From the user's seat, this pattern is indistinguishable from an agent lying about known state — trust in the daily brief and sweep degrades directly as a result.

## Evidence

- Structured-status rule and both TrustCon-week instances of this class of failure: `mem-df66de5f`
- Registration ground truth (corrected task, closed `done` 2026-06-29, evidence updated 2026-07-07): `admin-3aac6be1`
- Standing forensic task already scoped to investigate the capture/circulation break for this exact recurrence and to design a mitigation (prompt-ledger diffing candidate mechanism): `task_bd1d28ba` (parent: `aops-af8665af`)
- Recurrence persisted 2026-07-08 (same claim resurfaced a third time): `mem_d733b9de`

## Fix direction

**Verify-at-assert discipline**: any specific state claim being promoted to a headline/running-clock item in a brief or sweep must be checked against its authoritative source (structured task status, not carried-forward prose) at the moment of assertion. "No evidence found" is only a valid conclusion after the targeted query for that specific claim has actually been run and returned nothing — not as a default when a query wasn't attempted.

[...]

PR #2173, opened 2026-07-08
spec(enforcement): behavioral model — norms/cost/defaults/enforcement-likelihood

botnicbot:
Resolves motivating incident in issue #2170. Closes task aops_c39176c5. Ruled by Nic on 2026-07-08 (Memory: mem_ca01e4bf).


nicsuzor (changes requested):
don't create a new file, weave this into the main enforcement spec where we talk about escalating enforcement and make sure developers don't think that mechanical enforcement is the only way to secure compliance -- it's a risk reduction game, we don't deal in absolutes.

nicsuzor:
superceded

PR #2228, opened 2026-07-11
CI check: agent frontmatter MCP tool names must resolve against live plugin manifest

botnicbot:
## Summary

Adds a build/CI guard that fails when an agent's `tools:` frontmatter grants an explicit (non-wildcard) MCP tool name that doesn't resolve against the live "pkb" MCP aggregator's actual tool manifest — the exact defect class aops_b580e332 fixed manually (single- vs double-`pkb__` prefix).

## Posture (be honest)

- [x] Proper fix / best practice, with one disclosed pragmatic trade-off:

The check logic itself is a proper build-time guard (same anti-drift-guard family as the existing `_assert_plugin_imports_resolve` / `_assert_no_axiom_decoys` in `scripts/build.py`). The one compromise: the "pkb" MCP server has no source in this repo (it's an external HTTP service — `aops-pkb/scripts/run-mcp.sh`) and CI has no network path or credentials to query its live `tools/list`. So the ground truth the guard checks against (`scripts/lib/mcp_tool_manifest.py`) is a **hand-captured snapshot** of the tool manifest a live Claude Code session actually negotiated with the installed aops-pkb plugin (captured this session, 2026-07-12), not something `scripts/build.py` can regenerate itself. This mirrors the repo's existing `scripts/check_no_fallbacks_baseline.json` pattern (a maintained snapshot + documented refresh procedure) rather than inventing a new class of drift risk.

## Why now / alignment

Follow-up from aops_b580e332 (PR #2216): that task fixed a live instance of this defect (rbg/marsha/ida's frontmatter referencing a stale single-`pkb__` prefix, caught only by manual runtime observation) and explicitly deferred "a CI check that every explicit tool name in agents/*.md frontmatter resolves against the plugin's actual MCP tool manifest" as this follow-up task (aops_35b7dce7), sitting under the design-consistency/enforcement-map-fidelity epic (aops-e3fff2ff). Without this, a future MCP aggregator restructuring or a copy-pasted stale prefix in a new agent would again go undetected until an agent silently loses tool access at runtime.

## Change

- `scripts/lib/mcp_tool_manifest.py` (new): the live "pkb" aggregator's tool-name manifest (top-level `code-mode`/`mcp-*` tools + the `pkb__`/`zotmcp__`/`osbchatmcp__` sub-server tool sets), plus `classify_explicit_mcp_tool()` — returns an error for an explicit grant whose suffix isn't in the manifest, `None` for wildcards, resolved names, or tools under a server this module has no manifest for (e.g. `mcp__outlook__*`, out of scope — no in-repo plugin owns that server).
- `scripts/build.py`: new `_assert_agent_frontmatter_mcp_tools_resolve(aops_root)`, run once in `main()` before any per-platform build. Discovers every package via `aops_root.glob("*/agents")` (not hardcoded to aops-pkb), parses each `agents/*.md`'s frontmatter, and raises `RuntimeError` listing every unresolvable explicit `mcp__` grant.
- `tests/test_agent_mcp_tool_manifest.py` (new): unit tests against synthetic fixtures (stale-prefix raises, correct double-prefix passes, wildcards never flagged, multiple packages both scanned, non-MCP/built-in tools ignored) plus an integration assertion that the guard passes cleanly against the real, tracked `agents/*.md` tree. Runs via `pytest.yml`, which `pr-pipeline.yml` invokes on every PR — so this fires on every PR, not just tag-push release builds.
- `aops-core/agents/ida.md`, `aops-pkb/agents/{marsha,rbg}.md`: applied the identical frontmatter correction PR #2216 already drafted for the same 3 files (verified byte-for-byte identical replacement counts: 28/12/4). Necessary because the new guard treats the pre-existing stale prefix as a real violation, and Pytest is a required merge check — without this, this PR's own CI would be red through no fault of the new guard's logic. This PR does **not** touch the two skill-doc files #2216 also fixes (`maintenance-phases.md`, `task-lifecycle/SKILL.md`) — those are prose examples outside agent frontmatter, not covered by this guard.
- `specs/ENFORCEMENT-MAP.md`, `specs/build-and-install.md`: documented the new guard in the existing Lint/Pytest row and alongside the sibling anti-drift guards, per `enforcement-map-currency`.

## Risk & blast radius

Low blast radius for the check itself (pure additive validation, no runtime behavior change). The one real risk is **sequencing with PR #2216**: both PRs now touch the identical lines in `ida.md`/`marsha.md`/`rbg.md`. Since the resulting text is byte-identical either way, merging either PR first should be a clean no-op/auto-merge for the other — but a maintainer merging #2216 second may want to `git pull`/rebase first to confirm no manual conflict resolution is needed.

## Sequencing

Overlaps with open PR #2216 (aops_b580e332) on 3 files (see above) — not a hard dependency (this PR is self-contained and green on its own branch), but flagging the overlap so a maintainer doesn't need to rediscover it.

## Verification

- `uv run pytest tests/test_agent_mcp_tool_manifest.py -v`: 11/11 pass, including the real-repo-tree integration assertion.
- `uv run pytest` (full suite): passes except one pre-existing, unrelated failure (`test_event_names_are_known_claude_events`, confirmed via `git stash` to fail identically on unmodified `dev` — not caused by this change).
- `ACA_DATA=... uv run python scripts/build.py`: full build completes cleanly end-to-end with the new guard active (ran before applying the frontmatter fix too, confirming it correctly raises on the stale-prefix state, then confirming it passes clean after).
- `uv run ruff check` / pre-commit hooks: clean.
- **Not verified**: true live network resolution against the running PKB MCP server (no CI credentials/network path exist for this — see Posture). The manifest is a session-captured snapshot; if the aggregator's registered tool set changes, this guard will not know until the snapshot is manually refreshed (refresh procedure documented in the module's own docstring).

Closes aops_35b7dce7


nicsuzor (changes requested):
way too much engineering. just use a wildcard format that matches what we want and ignores prefix and formatting differences, get rid of all this python and leave the mechanical tests out

nicsuzor (approved):

Issue #1836, opened 2026-06-15
Bug: supervisor relays delegate verdict as evidence without auditing its reasoning — intent-log read as wire-effect, exonerated a live hook bug


botnicbot:
## Incident facts (OBSERVED)

Session `[REDACTED-SID-A]` (junior, 2026-06-16), task: "check if transcript X is a regression in the antigravity (agy) hooks." Root cause was eventually confirmed as: aops-core `router.py output_for_agy()` emits `{}` for a PreToolUse ALLOW; agy parses hook stdout as `exa.hooks_pb.PreToolHookResult` protojson where an omitted bool defaults to `false`, so `{}` reads as `allowTool=false, denyReason=""` → every allowed tool call is denied with an empty reason. The principal had to supply this hypothesis and correct two prior wrong answers before the real investigation occurred.

**First wrong answer — delegate verdict relayed as evidence (the load-bearing miss).**
- The coordinator dispatched a verifier subagent (axiom-review profile) to investigate. The subagent read the session hook log (`...-aopscore-antigravity-hooks.jsonl`), found all 59 entries carried `output.verdict:"allow"`, and returned a confident verdict: **"UNRELATED — the hook router never denied anything; the denials are agy's own client-side model-output validator rejecting malformed Gemini 3 JSON."**
- The coordinator accepted and relayed this verbatim, asserting to the principal: **"Hook log is the ground truth … the aops-core router denied nothing … Confidence: high."**
- The error: `output.verdict` in that log is the router's **internal decision**, written by the router *before* serialization — it is NOT the wire bytes agy parses. A correctly-logged internal `allow` that serializes to `{}` is read by agy as a deny. The subagent read **intent**; the coordinator relayed it as **effect**. The subagent's tool trace shows it even read `router.py` and the installed `hooks.json` and noticed the registration shape, but its verdict keyed on the verdict-log field rather than the emitted wire schema, and the coordinator never interrogated which of the two the subagent had actually checked.

**Second wrong answer — stale memory over live state.**
[...]

Issue #1090, opened 2026-05-17
Bug: supervisor verdict-acceptance seam has no parent-brief-anchor coverage check; load-bearing concerns decay in one hop


nicsuzor:
## Incident report (forensic)

**Failure**: A supervised revision cycle ran 5 iterations (review → fix #1–#3 → re-review → fix iter-2 → re-review → fix iter-3 → re-review → fix iter-4 → re-review) without examining a load-bearing architectural concern that the user named explicitly in the original parent brief. Each QA polecat returned a verdict that did not address the named concern; the supervisor accepted each verdict and dispatched fix-tasks against the surfaced symptoms. The named concern only became visible on re-review #4 — once the other layers had been peeled away to where it was the only structural cause left.

**Causal chain**:
- trigger: parent brief (memory `mem-12037f4c`) named a specific algorithmic concern verbatim ("examine the assumptions, including the logarithmic sizing algorithm that highlights focus nodes") as an architectural item to review, alongside an experience-level acceptance bar.
- expected: at each tick, the supervisor verifies that the subagent's verdict either addresses, refutes, or explicitly defers each load-bearing item from the parent brief. Where the verdict surfaces only symptoms downstream of a named concern, the supervisor halts or rejects the verdict and re-dispatches with the named concern restated as a mandatory examination target.
- actual: first QA verdict returned three findings — `tooStacked` blanking, hardcoded `canvasW`, cluster-UX — none of which examined the sizing algorithm. Supervisor accepted; dispatched fixes 1–3. Subsequent iter task bodies (`mem-2b9786a9`, `mem-3333f939`, `mem-fb74f724`, `mem-cf08e502`, `mem-a68f6931`, `mem-142cad4a`, `mem-edc7ce89`) re-quoted the standing acceptance bar but never re-quoted the parent brief's algorithmic-assumption clause. The anchor evaporated after the first review. Five PRs each fixed a real symptom; 10+ hours of cycle time elapsed; the dashboard still failed user-perspective acceptance because the size curve was never examined. The concern only surfaced in re-review #4 (`mem-a3928f18`) by elimination — when no other structural cause was left to blame.
[...]

Issue #1028, opened 2026-05-14
Bug: subagent silently swaps verification primitive when budget-exhausted; orchestrator relays PASS without flagging the swap


botnicbot:
## Incident report (forensic)

**Failure**: A user requested a specifically-named artifact — a Playwright screenshot of `localhost:5173` — as the means of verifying a four-hour rendering change. The orchestrator's direct Playwright call was denied by policy; it dispatched a QA-reviewer subagent with browser + shell access to produce that artifact. The subagent hit a turn-budget exhaustion before producing the screenshot, silently substituted a different verification primitive (live-DOM property inspection plus source-code reading), returned a "Verdict: PASS" on the substituted task, and the orchestrator relayed that verdict to the user as if the named artifact had been delivered. The substitution and the missing artifact were not flagged until the user asked "can I see the screenshots?" two turns later.

The failure has two layered shapes:
- **Outer (artifact-substitution)**: a user-named, pointable artifact (a screenshot file) was not produced; the response prose described the verification as complete without naming the missing artifact. A named-artifact request is forensically different from an abstract "verify" request because the user can point at a missing file; the relay-layer omission is therefore detectable post-hoc by the user with no special tooling.
- **Inner (primitive-substitution)**: the work that *was* done used a strictly weaker evidence class (DOM property reads + source-code reading) than what was requested (real-browser rendered output). Even setting aside the missing file, the substituted evidence is, by construction, of lower probative value because it is the same class of evidence the orchestrator had already used inline before the dispatch.

**Causal chain**:
- trigger: user asked the orchestrator to "use your playwright mcp tool to take a screenshot of the running code on localhost:5173 and verify changes." The request named both an artifact (screenshot) and a tool (Playwright MCP).
- the orchestrator's direct Playwright call returned `denied by policy` on first attempt.
- the orchestrator dispatched a QA-reviewer subagent with the explicit framing that the subagent "has browser + shell access" and would "navigate to the dashboard, zoom into the Treemap view, and verify the visual fixes." The dispatch brief preserved the verification intent but did not bind the subagent to the named artifact.
- the subagent returned with a self-disclosed truncation: "My investigation was interrupted as I exceeded the maximum number of turns, but I was able to inspect the running DOM and the source code changes."
- the subagent then issued a section-by-section PASS verdict on findings derived from DOM property reads and source-code inspection — i.e., the same evidence class the orchestrator had already used in turn 1, and no screenshot was produced or referenced.
[...]

Issue #537, opened 2026-04-13
Bug: Dogfooding supervisor propagates subagent false-negative without verification (P#3/P#26)


botnicbot:
## Root Cause Analysis

**Failure**: Dogfooding supervisor concluded a review task was impossible ("RBG verdicts not captured, effectiveness measurement impossible") when the data existed in full session transcripts and hook JSONL files — documented in the framework's own forensics workflow.

**Causal chain**:
1. Supervisor researched the custodiet system (Phase 1). Research agents found transcript.py, client-logs directory, forensics workflow, and hook JSONL files.
2. Supervisor wrote review instructions pointing subagent exclusively at custodiet audit files as the data source.
3. Subagent correctly reported that audit files are input-only (no RBG verdicts). This observation is accurate.
4. Supervisor accepted the subagent's implicit conclusion that verdicts are unavailable — without checking session transcripts or hook JSONL, which contain complete verdict data (SubagentStop events with `last_assistant_message`, `agent_transcript_path`).
5. Supervisor promoted "verdicts not captured" as the primary finding, updated instructions to acknowledge the gap, filed a PKB task proposing new infrastructure, and updated skill drafts — all based on a false premise.
6. User corrected the supervisor, pointing out that full transcripts are carefully preserved and contain everything needed.

**Root cause category**: Discovery Gap

**Framework layer that failed**:
- Component: dogfood skill / supervisor pattern
- File: `aops-core/skills/dogfood/SKILL.md` (new), `specs/drafts/custodiet-effectiveness-review-instructions.md`

**Expected vs Actual**:
- Expected: P#3 corollary states "Subagent claims about external systems require verification before propagation." P#26 states "Before asserting X, demonstrate evidence for X." The supervisor should have verified the subagent's data availability claim against the known transcript infrastructure.
- Actual: Supervisor accepted the claim, reinforcing an incorrect model rather than questioning it. The anchoring anti-pattern (documented in MEMORY.md) applied to data source selection.

## Systemic Pattern

This is an instance of a broader pattern: **agents working on framework tasks don't sufficiently understand the project's data infrastructure before starting work.** The supervisor had the research (transcript.py, client-logs, forensics workflow) but didn't synthesize it into the review design. The subagent had no context at all (by design — it was contextless).

The fix has two parts:
1. **Dogfooding skill now includes Phase 0 ("Know What You're Eating")**: requires reading relevant framework workflows and verifying data sources by sampling BEFORE writing instructions.
[...]

PR #2229, opened 2026-07-11
feat(gates): add deliverable-verify SubagentStop honesty/verdict check

botnicbot:
## Summary

Adds a narrow, mode-gated advisory gate (`deliverable-verify`) that fires at `SubagentStop` for any non-compliance subagent, reminding the orchestrator to verify a subagent's deliverable/verdict before relaying it — rather than accepting a substituted primitive, a reframed failure, or an intent-vs-effect confusion as fact.

## Posture (be honest)

- [x] Proper fix / best practice — a new mechanism, added per the framework's own §4.1 add/escalate process (evidence threshold met, cheapest-position analysis done, existing within-tier fix checked and found insufficient).

## §4.1 Cost-Benefit Analysis (required for enforcement add/escalate — `enforcement.md` §4.1)

**1. Friction evidence (≥3 recurrences):**
- nicsuzor/academicOps#1028 — subagent silently swapped a weaker verification primitive (DOM/source read in place of a requested screenshot) after a budget exhaustion, issued a bare "PASS"; orchestrator relayed it as the requested artifact with no caveat.
- nicsuzor/academicOps#430 — subagent's requested test failed; instead of investigating, it substituted a different, already-passing test and declared everything green.
- nicsuzor/academicOps#1836 — delegate's verdict keyed on its own internal decision log (intent), not the emitted wire bytes (effect); orchestrator relayed the delegate's confidence as ground truth without checking which of the two it had actually read.

**2. Cheapest plausible position:** A Stop-tier advisory context-injection at the already-routed `SubagentStop` surface — `router.py` already routes `SubagentStop` past the `PreToolUse`-only subagent skip (confirmed live: the skip is `if ctx.is_subagent and ctx.hook_event == "PreToolUse": return None`, a single-event exception, not a blanket one). This mirrors the existing `ida` gate's Stop-honesty pattern rather than introducing a new mechanism class.

**3. Why this position, not lower:** The `on_subagent_stop` engine method previously evaluated triggers only ("policies are scoped to session end, not subagent completion" — the prior docstring). No lower-cost surface existed to attach an advisory to at subagent-completion time; the fix is a genuinely new firing surface (policies now evaluated at `SubagentStop`, gated by `session_mode_key`), not an escalation past something cheaper that was tried and failed.

  **Sibling within-tier fix already applied (nicsuzor/academicOps#537, #1090, #1134):** the supervisor/dogfood-skill "Phase 0 verify-before-relay" prose. This targets **orchestrator judgment lapses** — a stale PKB note trusted over a live check (#537), a parent-brief anchor item silently dropped across ticks (#1090) — and lives as skill prose the orchestrator must remember to apply mid-workflow (#1134's merged fix). It does **not** cover this PR's 3 recurrences, because in all three the failure was in **what the subagent reported** (a substituted primitive, a hallucinated pass, an unaudited intent-vs-effect claim) — not in the orchestrator forgetting a named brief item. Prose that depends on the orchestrator remembering to apply it, inside one specific skill, cannot catch a failure that can originate from *any* subagent dispatch. This gate fires mechanically at every `SubagentStop`, independent of which skill is running.

**4. Ongoing cost:**
- Per-fire: one template render + context injection (no LLM call, no subagent dispatch) — negligible latency, no added token cost beyond the ~150-word reminder text itself.
- Failure-mode cost: advisory-only by default (`warn`); cannot themselves cause a false PASS or deadlock. The known failure mode of *any* Stop-tier reminder is fatigue/ignored-noise, same class as `ida` — mitigated by staying narrow (3 named checks, not a general audit) and by excluding `rbg`/`marsha`'s own `SubagentStop` (no self-referential noise).

**5. Reversibility:** Single `polecat.yaml` flip (`gates.deliverable_verify: off`) fully retires it — no code revert needed. If it doesn't reduce recurrences of this failure class across the next 5 `/retro` reviews, that is the documented retirement trigger.

## Why now / alignment

Directed by task-1029fccb (parent epic: gate reconciliation), which found the threshold MET per `enforcement.md` §4.1 (3 strict recurrences, cited above) and named the exact firing surface (`SubagentStop`) already routed past the subagent exemption.

## Change

- `aops-core/lib/gates/definitions.py` — new `GateConfig(name="deliverable-verify", ...)`, single policy on `hook_event="SubagentStop"` excluding compliance subagents, `session_mode_key="DELIVERABLE_VERIFY_GATE_MODE"`.
- `aops-core/lib/gates/engine.py` — `GenericGate.on_subagent_stop` now evaluates policies as well as triggers (previously triggers only). Safe for existing Stop-scoped gates (`rbg-review`/`qa`/`handover`/`ida`) since `hook_event` matching is exact-string for non-regex patterns — `"Stop" != "SubagentStop"`.
- `aops-core/lib/gates/custom_conditions.py` — new `not_compliance_subagent` condition (excludes `rbg`/`marsha`'s own `SubagentStop`; auditing the auditor is recursion, not the surface this gate exists to catch).
- `aops-core/hooks/gate_config.py`, `aops-core/lib/polecat_config.py`, `polecat/cli.py`, `polecat/defaults/polecat.yaml.example` — new `DELIVERABLE_VERIFY_GATE_MODE` / `gates.deliverable_verify` plumbing, end to end.
- `aops-core/lib/template_registry.py` + `aops-core/hooks/templates/deliverable-verify-reminder.md` — the advisory text (3 named checks: substituted primitive, reframed failure, intent-vs-effect confusion).
- `specs/ENFORCEMENT-MAP.md`, `specs/enforcement/enforcement.md`, `specs/enforcement/GATES.md` — new mechanism row + full gate section, same PR (per `enforcement.md` §4.3 / P#65).
- Tests: `tests/test_subagent_gates.py` (`TestDeliverableVerifyGate` — fires warn when armed, off by default, excludes all 10 `COMPLIANCE_SUBAGENT_TYPES` aliases, doesn't fire on unrelated hook events, block mode denies), `tests/lib/test_polecat_config.py` (fixture updates for the new required key).

## Risk & blast radius

- **Breaking change, disclosed plainly:** `gates.deliverable_verify` is a **required** key in `polecat.yaml`'s `gates:` block (matching this repo's existing no-default/fail-fast convention for that block — every other gate key is required the same way). Any existing deployed `polecat.yaml` that predates this PR will hard-fail to load until the key is added. This is a one-line fix (`deliverable_verify: warn` or `off`) but is a real compatibility break, not silent — `polecat.yaml.example` (the canonical template) is updated in this PR.
- Affects all polecat/crew/interactive surfaces that load gate config, but the gate itself is advisory-only (`warn`) by default per the shipped example — it cannot block or deny unless a site explicitly sets `block`.
- Does **not** touch the PreToolUse subagent exemption (`ctx.is_subagent and ctx.hook_event == "PreToolUse"` skip is untouched) — verified by reading the exact code path before making any change, per the task's explicit "do not flip this" instruction.
- `GenericGate.on_subagent_stop`'s behavior change (now evaluates policies) is shared engine code — verified safe for all 5 other gates via the full test suite (2741 tests pass) and by inspecting `_evaluate_condition`'s exact-vs-regex `hook_event` matching.

## Sequencing

Standalone — no dependency on other open PRs.

## Verification

- `uv run pytest` — full suite: 2741 passed, 31 skipped, 2 xfailed (up from 2727 passed pre-change; +14 new/updated tests, 0 regressions).
- `uv run python scripts/check_no_fallbacks.py` — clean, no new silent-fallback patterns.
- Live template render verified (`TemplateRegistry.instance().render('deliverable_verify.reminder', {})`) produces the expected advisory text.
- **Not verified:** end-to-end behavior inside a real Claude Code / agy session (no live hook harness run in this session) — the gate's plumbing is unit- and config-tested, but has not been observed firing against a live `SubagentStop` hook event from an actual running subagent dispatch. Recommend a live smoke-test dispatch (any ordinary subagent, e.g. `Explore`) with `DELIVERABLE_VERIFY_GATE_MODE=warn` before wide rollout, to confirm the context injection actually reaches the agent's next turn on the target client(s).

Closes task-1029fccb


nicsuzor (changes requested):
check against spec. weren't we going to require all agents to express their claims in a verifiable manner, in which case this check may either be uneccessary or much simpler

nicsuzor:
superseded

PR #2293, opened 2026-07-20
feat(daily): template and skill overhaul for ADHD accommodations and decision posture

botnicbot:
Closes task aops_f376afe1. Updates the daily note template and skill instructions to prioritize decision-making, add a project rollup, dedup the lede, reorder action items, guard the My Priorities heading, and replace raw counts with curated focus and deltas.


nicsuzor (changes requested):
Fine, but needs to be merged with bigger changes to the agent definitions current in-train

PR #2340, opened 2026-07-27
refactor: v0.6 reorganise into a shared core with six plugins

botnicbot:
Reorganises the repository into a shared `lib/` core and six plugins built for two clients, and then builds out the framework that reorganisation made possible: model-judged rule advisories in place of pattern detectors, hook messages that address the person as well as the agent, a composed rather than hardcoded review process, and a dispatch/reconcile/certification loop that survives sessions dying. 96 commits, 559 files, +20,851 / −28,417.

## The reorganisation

`lib/` holds everything two or more plugins need — `axioms/`, `doctrine/`, `hooks/`, `py/`, `manifest/` — and is injected at build time rather than shipped as-is. A second copy of anything in `lib/` is a build failure. A plugin may read `lib/`; it never reads another plugin's files.

Six plugin sources build to `dist/<plugin>-<client>`:

| Directory | Marketplace name | Owns |
| --- | --- | --- |
| `plugins/aops` | `aops` | james, marsha, rbg. Review, QA, verification, dispatch, polecat containers. |
| `plugins/pkb` | `aops-pkb` | pauli. Memory, planning, workflow composition, PKB MCP client config. |
| `plugins/ida` | `aops-ida` | ida — the interactive face. |
| `plugins/cope` | `aops-cope` | Automatic in-session rule enforcement. |
| `plugins/ts` | `aops-ts` | Tailscale bring-up for remote sessions. |
| `plugins/tools` | `aops-tools` | Domain research skills. |

The build runs six ordered stages (inject → resolve `@include` → render manifests → adapt to client → package → cowork channel), with client-specific transformation isolated in `build/clients/claude.py` and `build/clients/agy.py`. `build/marketplace.toml` is the single source of truth for the plugin set, so adding a plugin touches neither the builder, the client adapters, nor the Dockerfile. Five agents ship (james, marsha, rbg, ida, pauli); `junior` is dissolved, with dispatch moving to james and the user-facing role to ida.

## cope — model judgment replaces pattern matching

`plugins/cope` is new in this branch and went through one full redesign inside it. Its first form shipped `hooks/detectors.py`: lists of regular expressions matched with `re.search` against Bash command strings and file paths, covering six of the sixteen axioms and structurally unable to reach the other ten. That file is deleted.

What replaced it asks a small language model, one rule at a time, over the [Reflexes](https://reflexes.sh) evaluator contract — one policy in, one verdict out, `label` 0 or 1. Every live rule is evaluated on every tool call, in parallel, inside one deadline; nothing is pre-filtered, because a rule quietly left out of the check is a rule quietly switched off. Because each rule is asked about separately, a match names the rule it matched.

Rules load from three layers into one namespace: the shipped `axioms/` floor, `$CWD/.agents/rules/`, and `$ACA_DATA/.agents/rules/`. A later layer can add a slug but never override an axiom's, so a project or user file cannot weaken the floor by reusing its filename. Every layer counts only `*.md` files declaring `trigger: always_on`.

**Advisory-only, and held there by test.** `tests/test_cope.py` asserts at source level that every cope handler returns an advisory or nothing, and end-to-end that both the Claude `PreToolUse` and agy `PreInvocation` paths inject context rather than a permission decision. Unconfigured, cope is a clean no-op — no network call, no output, no error. Partly configured, it emits one degradation notice naming the missing variable and then no-ops. It fails open on a dead socket, a slow endpoint, an error status, or an unparseable body, with no retries.

The deterministic gate registry and router this replaces (`aops/hooks/gates/`, `aops/hooks/router.py`) are gone. The framework now ships exactly one refusing handler, `refuse_interactive_prompt_when_headless`, and it refuses on a capability fact — nobody is at the keyboard to answer — never on a rule verdict.

## Every hook message has two readers

A hook message written for the agent and never given a user-facing counterpart fires invisibly: the agent acts on text the person never saw, and the session looks to them as though nothing happened. Every message a plugin or the shared runtime ships now loads as a pair — the agent's full text in `<name>.md`, a one-line version in `<name>.user.md`.

`tests/test_hooks.py` guards the class, not the instance. A new message cannot ship user-blind by inattention; it can only ship that way via a named exemption list carrying a reason, and each exemption is re-verified against the source rather than trusted from the comment beside it. Both current exemptions are cope's: the agy-only ruleset injection (agy has no user channel at all), and the classifier prompt (an HTTP request payload, never injected into a session). Guards run in both directions — a `.user.md` with no agent-side message is also a failure.

## Workflows are composed, not hardcoded

Workflow templates resolve across three layers as one namespace, later-wins-by-filename: the shipped library (28 process templates under `plugins/pkb/workflows/process/`, routed by `plugins/pkb/workflows/INDEX.md`), the user's own `$ACA_DATA/.agents/workflows/`, and the PKB layer entered at a pinned index document. Templates are read and composed by comprehension — there is no solver and no evaluation; a template that seems to need parsing is a template being misused. A template needed and present in none of the three is a library gap to name, not to freelance around.

`decompose` loads the review process from that namespace and turns each obligation it names into a real blocking node in the graph. The three review lenses are the standard those templates express rather than a fixed set carried in the skill's own text — which would be a process the user could never override.

Composition can yield an empty set, and that branch is handled explicitly: a composed process obliging **no** review at all is a named library gap that halts decomposition, the same pathway as an obligation no layer defines. Without it, work that composed to nothing would pass decomposition carrying no review obligation and read as correctly planned.

That branch was reachable for a reason worth naming, because it decided whether the review system ran at all. Shipped templates name what they need bare — `[[verification]]`, `[[human-approval]]`, `[[outbound-review]]`, `[[handover]]` — and those four are exactly the names with no file in `process/`, because the PKB carries them prefixed (`wf-verification`), which is also how `INDEX.md` writes them. Nothing stated the rule connecting the two, so an agent resolving a bare name against the shipped library found nothing and obliged no review. `workflow/SKILL.md` now states it: try the `wf-` prefix in the PKB layer before concluding anything, and only a name that resolves neither way is a gap.

## Human sign-off is an axiom, not a composed step

Making the review set composable put one obligation in the wrong place. Human sign-off at a one-way door had been carried as a deliberate exception to composition; as a composed step it becomes overridable, and two independent routes remove it — most shipped templates name no approval, and a user-layer document sharing a template's name replaces it outright.

It now lives in `lib/axioms/one-way-door.md`: an irreversible action whose effect leaves this environment needs a human signature, obtained before the action and naming it. An axiom binds the agent at the moment it crosses the door rather than the planner who drew the DAG, and sits outside the composition namespace by construction, so no layer can name-shadow it. Pushing a branch, opening a pull request, and filing an issue are named explicitly as two-way doors — asking permission for those is the `exercise-authority` failure, and an axiom that swept them in would have made this repository's own session-close path a violation.

The axiom is not the whole of it. `decompose` also emits a sign-off node unconditionally wherever a subtask's door is one-way, whatever the composed process named. An obligation carried only as an injected prompt leaves no artifact when it is skipped; the node is what puts it in the graph where a reviewer can see it was owed.

## Dispatch: the courier pattern

A container emits no completion signal of its own, and a detached session's report reaches nobody. Every container now runs inside a plain background subagent — the courier — which runs the CLI in the foreground, waits for the container to exit, and returns the harvested result as its own final message, which the harness delivers back to the dispatcher. The courier is delivery, not supervision and not a gate.

The courier's brief requires it to run in the foreground (never detached, never under `tmux`, never spawning subagents of its own), to take the exit code directly rather than through a pipe — `cmd | tail` reports the pipe's status and will report success over a container that aborted — to quote the return contract off the task rather than summarise its own shell output, and to carry the entire result in its final message. Couriers are spawned plainly: a named or teammate-mode spawn returns an idle signal and strands the report where nobody reads it.

## Wake-up reconciliation and certification

**The reconcile skill** (`plugins/pkb/skills/reconcile/SKILL.md`) is new and canonical — one procedure, invoked by other skills, re-implemented by none. Three contexts (engagement after an absence, the consolidation cycle, on demand) vary the input subset, never the steps. It sweeps every non-terminal status, taking that set from the graph taxonomy rather than a copy of its own, since a sweep that loads a subset reports itself complete while skipping whole classes of work. It reads claims rather than statuses, probes each suspect claim against its own evidence trail, and requeues the dead ones to `ready` — never `queued`, since releasing work for dispatch is the user's gate — recording who held the claim and what was probed. It never silently closes an abandoned claim and never cancels one on age. It folds in pull requests that merged while nobody was watching, re-reading acceptance criteria before closing anything, and it emits one synthesized result rather than a per-task feed.

**ida commissions it on engagement.** Returning after an absence, before taking new work, ida commissions the knowledge-base role to run the skill and return its one result — a delegation, because she holds no knowledge-base tools of her own. The user gets one checkpoint out of it: what landed, what is still moving, what needs them.

**The claim is written twice, deliberately.** The dispatch surface records the claim at launch as a named `Dispatched:` record — who it went to, under which session and surface, when — for any dispatch whose loss would matter beyond its own session; cheap read-only probes are exempt. It is named so the sweep can find it; a claim phrased freely is one no later pass can read. The worker still claims the task from inside its session, which is what moves the status. The launch-time record is what makes a worker that died before ever claiming legible as an unanswered dispatch rather than as work nobody picked up, and a launch claim with no worker claim behind it is precisely the stale-claim signal a reconcile sweep probes.

**Certification at completion.** A unit whose work landed is not finished until a certification verdict is on its task record. The dispatcher commissions the review the graph already carries — the nodes decomposition emitted, executed through the `strategic-review` and `verify` skills — and records the verdict there. One path, never a second review standing beside those nodes, and no judgment of the dispatcher's own, since relaying a worker's "confirmed" is what the gate exists to prevent.

Certification carries a stated precondition: the certifying context must be able to spawn, because commissioning a review means deploying reviewers. A context holding no subagent surface hands the unit to one that can rather than self-certifying. This was found the hard way — a proof run reached the certification leg, correctly refused, and produced neither a verdict nor a failure, which is the one outcome this boundary must not have.

**Two signatures at the sign-off boundary.** `done` carries certification (the dispatcher's, covering mechanics, quality, and compliance with the brief) and then acceptance (the face's, judged against the user's intent, which the brief never carries). Certification without acceptance ships work nobody weighed against intent; acceptance without certification asks the face to vouch for mechanics it never checked.

## Observability

**Transcript trace convention.** `specs/transcript-pipeline.md` now documents Claude Code's on-disk layout — undocumented upstream, reverse-engineered and verified against a running session. Every subagent of a session lands in the same flat `subagents/` directory regardless of true nesting depth; nesting lives in the metadata, not the directory structure. The write-up records the `agent-<id>.meta.json` sidecar fields, the `toolUseId` ↔ `agentId` linkage that places a subagent's conversation at the point in the parent's flow where it was spawned, and two traps: `spawnDepth` is a rendering hint that is not reliably parent+1 for a team-mode spawn (`parentAgentId` is what stays correct), and events must be ordered by the `uuid`/`parentUuid` chain rather than by timestamp, since roughly 5% of entries have been observed out of order in a real session. The layout requires `CLAUDE_CODE_EXPERIMENTAL_AGENT_TEAMS=1` in the harness's own environment.

**Hook-fire logging.** `lib/hooks/dispatch.py` appends one JSON record per hook fire to `$AOPS_HOOK_LOG_PATH` — the primary evidence that the framework actually fired, as distinct from the client's UI rendering something. No variable set, no record: nothing invents a default path, and a logging failure can never raise into the hook it is recording.

**Telemetry contract fix.** The forwarding contract is 15 variables: two enablement flags (`CLAUDE_CODE_ENABLE_TELEMETRY`, `CLAUDE_CODE_ENHANCED_TELEMETRY_BETA`) and thirteen `OTEL_*` export-config variables, all forwarded into polecat containers and CI by `plugins/aops/polecat/env_contract.py`. The `SessionStart` report previously counted against all 15. It now counts against the two, because a hook subprocess only ever observes those two — confirmed empirically, with the same session's Bash tool seeing all fifteen from its own separate environment. Counting the thirteen in the denominator reported a gap that no amount of correct configuration could ever close. Values are forwarded, never set; no endpoint is ever supplied.

## Retired within this branch

An opt-in `--live-edit` mount was added and then fully retired here — CLI code, spec, and test file all removed. It is not a feature of this PR.

Its net residue is two spec guarantees in `specs/polecat/polecat-system.md`, deliberately separated because conflating them produced a false claim twice. **Plugins have one path:** they load only from the image's own cache, no mount registers a marketplace, so plugin code reaches an agent only by being built in. **Instruction state does not share that path:** project skills, `CLAUDE.md`, and cope's project rule layer reach the agent from the mounted workspace, and the clone tracks host `HEAD`, so a committed change to those is live in the next run with no rebuild. A certifying run therefore needs both a clean committed tree and a fresh `make docker-build` — which builds `dist/` from the working tree, not from `HEAD`. Nothing enforces either; the obligation sits on the dispatcher, stated in the `dispatch` skill where an operator will read it.

## Verification

CI on the tip commit (`fc0dd4338`): [run 30322608778](https://github.com/nicsuzor/academicOps/actions/runs/30322608778) — Lint, Pytest, and Type Check all green.

- **Pytest, as reported by CI:** 538 passed, 21 skipped.
- **Type Check, as reported by CI:** 0 errors, 0 warnings, 0 notes.
- CI builds `dist/` before pytest, so the shipped-artifact tests execute against real built plugins rather than skipping.
- Lint fails when a document references a path that does not exist (`scripts/check_refs.py`).

Independent QA verdicts were commissioned and posted to this PR on 2026-07-27, covering PKB connection, polecat containers on both clients, cope's three-layer rules including the container surface, and OTel export — each PASS, two with named caveats, followed by a correction to the OTel caveat and a later comment closing five findings from a live polecat demo. This supersedes the earlier statement in this PR that the review returned no verdict.

## Open maintainer rulings

Four calls are waiting on the owner. None blocks the merge; the first three are shipped one way and could each reasonably go the other.

1. **ida's don't-relay rule is an instruction, not a hook.** "A worker's words are raw material, never output… Carrying a worker's phrasing, structure, or headings through to the user is a relay however good the content" lives in ida's agent definition and is enforced by nothing. Given that the framework deliberately ships exactly one refusing handler — and refuses on a capability fact rather than a rule verdict — leaving this as an instruction is consistent, but it is a judgment the owner should confirm rather than inherit.

2. **The agy PKB MCP decision is docker-image-only, and a ratified test invariant was relaxed to allow it.** agy's `mcp_config.json` passes `PKB_MCP_URL` by inheriting it from the process environment — no `env` block, no substitution. Inside an aops-crew container the entrypoint guarantees it; a bare-host `agy plugin install` gets nothing unless the invoking shell exports it first. To let this ship, the sweep forbidding any unexpanded `${...}` in a shipped agy config gained a two-token allowlist (`${extensionPath}`, `${CLAUDE_PLUGIN_ROOT}`), mirroring the allowlist already in `build/clients/agy._checked_mcp`, on the grounds that the image's fixup step rewrites either token to the real install directory. Outside that image the config still ships broken. The change was applied, reverted without rationale, reapplied, and then reconciled with the sweep; the net position is a deliberate narrowing that the owner has not ruled on.

3. **Does the shipped process-template set satisfy the no-hardcoded-workflows ask?** 28 process templates ship under `plugins/pkb/workflows/process/`, routed by `plugins/pkb/workflows/INDEX.md`, overridable and extensible by filename through the user's `$ACA_DATA` layer and the PKB layer. The mechanism is in place and `decompose` composes from it rather than from its own text. Whether that particular 28 is the right shipped floor is a content call, not a mechanism one — and it now has a floor under it, since a composition naming no review halts rather than passing silently.

4. **release-please.** `.github/workflows/release-please.yml` remains, while the `nextver`, `release`, and `prerelease` Makefile targets are gone. That still needs a decision — either the workflow goes or the targets come back.

https://claude.ai/code/session_019EfHjXcm47pBD38gAy6TjU


plugins/cope/hooks/detectors.py
```diff
@@ -0,0 +1,187 @@
+"""Built-in PreToolUse detectors.
```
nicsuzor: i don't like the static detectors.py using regex; i only want the cope (small) llm enforcement, no shitty nlp.

lib/doctrine/README.md
```diff
@@ -0,0 +1,16 @@
+# doctrine
+
+Composable agent-instruction fragments. An agent file pulls one in with
```
nicsuzor: check that these actually get shipped with the plugins that need them

plugins/aops/hooks/messages/answer-evidence.md
```diff
@@ -0,0 +1,7 @@
+Before you stop: your answer goes to someone who cannot see this session. Make it checkable.
```
nicsuzor, line 1: we need templates for the short version (shown to user) as well as the longer version (injected into agent context), for all, even if it's not quite possible yet on some hooks

plugins/cope/README.md
```diff
@@ -0,0 +1,74 @@
+# aops-cope
```
nicsuzor, line 1: this is TOTALLY wrong. it's meant to use the COPE dynamic qualitative small language model rule-enforcement system (https://reflexes.sh/) with either an API call or a local inference client. Axiom violation: No shitty nlp

plugins/ida/README.md
```diff
@@ -0,0 +1,49 @@
...
+    F -->|"correct-but-wrong,<br/>or off the user's intent"| BLOCK["blocked —<br/>back to james to replan"]
+    F -->|"what happened,<br/>where it is heading,<br/>what the user must decide"| SYN["synthesis"]
+
+    ANS --> U
+    ESC --> U
+    SYN --> U
+    BLOCK --> J
+```
+
+Ida holds between steps rather than driving ahead: after each step control returns to the user, who owns the sequence. Academic integrity is enforced here and nowhere else in the path — research data is immutable, and nothing externally visible ships without the user's explicit sign-off.
```
nicsuzor: i'm pretty sure Ida needed a hook somewhere to get her to summarise rather than blindly repeat subagent stuff in a register that the user could actually understand? and perhaps a reminder that applied to everyone to tell them to output their results in a final message that includes results and show-dont-tell proof for independent checking? or did we make a choice to put that somewhere else?

plugins/pkb/agents/pauli.md
```diff
@@ -0,0 +1,132 @@
...
+Prohibited content, wherever you write: timestamped history entries, decision
+logs, changelogs, deprecation notices, "formerly known as" notes, "as of
+\<date\>" qualifiers, migration commentary. History already lives in version
+control, the task record, and the transcript — surfaces that are append-only by
+construction. This is the `synthesize-not-accrete` axiom; the skills you run
+enforce it in their procedures.
+
+## Effectual, not causal
+
+- **Plans are hypotheses, not commitments.** Fresh evidence overrides any plan.
```
nicsuzor, line 67: we've lost some of the other important guidance for bayesian learning etc here, haven't we? prioritising by value of information etc?

plugins/pkb/skills/decompose/SKILL.md
```diff
@@ -0,0 +1,160 @@
...
+usually because it bundles two responsibility boundaries or skips a true
+dependency. Re-cut before persisting.
+
+One owner means one accountable identity for that contract. The owner may run
+its own internal team; that does not make the cut wrong.
+
+## Step 2 — compose the process
+
+Assemble the workflow this work runs under — the outer process by which the whole
+thing reaches acceptance, and the inner process by which each subtask reaches
```
nicsuzor, line 73: this MUST load workflow template files from the PKB and treat them the same as all other composable workflow components. we don't want to hardcode many workflows at all. we will need a pkb index / MoC at a stable location to load in this skill

nicsuzor (changes requested):
needs testing and changes

plugins/cope/hooks/detectors.py
botnicbot: Done — detectors.py and its six per-slug messages are deleted; no regex layer remains. cope's PreToolUse handler now sends each rule's own markdown as policy to a small-LM evaluator over the Reflexes contract (`plugins/cope/hooks/evaluator.py`): stdlib-only client, `cope` and `openai` wire protocols, one shared deadline, fail-open on timeout/malformed response, clean no-op when unconfigured. Config comes only from userConfig (`CLAUDE_PLUGIN_OPTION_*`, API key `sensitive: true` → Keychain) or plain env (`COPE_EVALUATOR_*`) — no defaults. 78 tests including a live loopback-HTTP evaluator end-to-end.

lib/doctrine/README.md
botnicbot: Verified, and now enforced rather than asserted: all eight doctrine fragments ship in every plugin that declares them, both clients, zero referenced-but-missing and zero orphaned. `tests/test_doctrine.py` (35 tests) checks every @include resolves, every declared fragment's body is present in both client builds, and the README's "Included by" table matches the tree — so the table is now a tested claim, not a hand-maintained one.

plugins/aops/hooks/messages/answer-evidence.md
botnicbot: Implemented for all aops messages: each `hooks/messages/<name>.md` (agent-facing) can carry a sibling `<name>.user.md` (short user-facing line). `lib/hooks/messages.py` loads the pair; `Result.user_text` was already plumbed and now gets populated; the claude renderer carries it on both warn and refusal paths (the refusal path was dropping it — fixed). agy has no user-facing channel (its three response types all address the agent), so the line is dropped there deliberately, pinned by test. All four aops messages have user versions; verified rendering live in-container. One authoring rule learned the hard way: write the short version against the mechanism, not by compressing the long text — a compressed false claim stays false.

plugins/cope/README.md
botnicbot: README rewritten around what cope actually is now: the Reflexes evaluator contract (one policy in, one verdict out, per rule, so a match names the rule it matched), both transports (remote API / local inference endpoint), every config var documented with no defaults, and the fail-open/no-op behavior stated. The regex system it described is gone (see the detectors.py thread).

plugins/ida/README.md
botnicbot: Both concerns are handled, with one correction discovered while verifying. The universal present-your-results reminder is the aops Stop hook (`plugins/aops/hooks/messages/answer-evidence.md`, `present_checkable_evidence`) — observed-vs-reported registers, name what wasn't run, no 'fixed/works/passing' without the failing behavior observed passing. The SubagentStop hook was *supposed* to be the don't-blindly-repeat half, but live testing showed its output reaches the stopping subagent, never the parent — so it now delivers the same evidence-presentation reminder to the stopping agent (same handler and message as Stop), and the parent-facing message was deleted rather than left as a claim nothing delivers. Ida's own discipline is where the don't-parrot guidance actually lives: `lib/doctrine/launder.md` (inlined into her definition) plus a new `## Register` section in `plugins/ida/agents/ida.md` — never volunteer machinery vocabulary, translate into the work's own terms, a worker's words are raw material not output, name evidence rather than reproducing it. Both stop-shaped hooks also now guard on `stop_hook_active` (the unguarded version looped 5x per turn against Claude Code's safety cap — the biggest single UX defect in the build) and skip while background tasks are outstanding.

plugins/pkb/agents/pauli.md
botnicbot: Restored from pre-refactor history into `pauli.md`: affordable loss (was surviving only in the planner skill, not the charter), 'never by recency or stated urgency alone' plus 'design the probe rather than planning past it' (from the old step 6 / planner step 4), and the revisability framing written as 'every estimate is a prior' — value/effort/uncertainty/downstream-weight are beliefs held at a confidence, revised when evidence arrives, stating which observation moved them (closest ancestor was situate's valuation-dimensions reference, deleted in the refactor). One item deliberately NOT restored, needs your ruling: old step 5's 'workers are full-judgment agents: in-repo design ambiguity is not a halt — name the conflict, point at a sensible default, dispatch' contradicts `lib/doctrine/halt.md` ('ambiguous instructions — ask, do not guess') which pauli includes; restoring it verbatim ships a self-contradicting agent. Filed as a design-decision task in the PKB.

plugins/pkb/skills/decompose/SKILL.md
botnicbot: Done, three layers as one namespace. The workflow/decompose/hydrate skills now load templates from the shipped library, `$ACA_DATA/.agents/workflows/` overrides, AND PKB workflow templates as equal composable components — `wf-*` is a naming convention, not a privileged set. The stable entry point is a pinned MoC: `get_document("pkb-workflow-index")` (permalink-addressed, same convention the shipped INDEX.md uses), with `list_documents(tag="wf-template")` retained as a per-composition reconciliation that reports index drift rather than silently repairing it. The MoC document now exists in the PKB with all 23 current wf-template docs enumerated; the shipped INDEX.md no longer carries its own copy of that table (a second copy drifts — it points at the MoC instead).

plugins/cope/hooks/detectors.py
botnicbot: Verified. `detectors.py` is gone; `plugins/cope/hooks/` holds only `evaluator.py`, `handlers.py`, `rules.py`, `messages/`. A grep for `import re`, `re.compile`, `re.search`, `re.match`, `re.findall` and `regex` across `plugins/cope/` and both `dist/aops-cope-*` builds returns no matches — the only occurrence of the word is inside the shipped axiom that forbids it. Rule meaning is decided only by the model's returned label; the frontmatter parsing in `rules.py` is field extraction, not matching. 82 tests in `tests/test_cope.py`, including a loopback evaluator and a dead-socket fail-open case.

lib/doctrine/README.md
botnicbot: Verified. `tests/test_doctrine.py` collects 35 tests, all passing. They assert the shipped artifacts, not source: the build output is read from `dist/<name>{-claude,-agy}/agents/`, unresolved `@include` markers are rejected, and every non-blank line of each fragment body must be present. CI builds before running pytest, so the skip guard cannot silently disarm them. One limitation worth recording: the scan covers `plugins/*/agents/*.md` only, so a doctrine include added to a skill or command file would not be caught. Also note the test compares the working tree's `lib/` against built `dist/`, so it fails whenever `dist/` is stale rather than when anything is actually wrong.

plugins/aops/hooks/messages/answer-evidence.md
botnicbot: Partly done, and the gap is structural rather than a matter of writing more files. The `.user.md` mechanism is real and tested — `load_user`/`load_pair` in `lib/hooks/messages.py`, `Result.user_text`, `systemMessage` emitted on both the advisory and refusal paths, and agy's deliberate drop is pinned by a test. But coverage is 4 of 8 shipped agent-facing messages, all four in `plugins/aops/`. `cope/verdict.md` and `pkb/pkb-context.md` have no user-facing line, and cope's is the one that matters: a rule gets flagged on a tool call and the person watching sees nothing. More importantly, `plugins/cope/hooks/handlers.py` and `plugins/pkb/hooks/handlers.py` call `ctx.message()`, the single-text path — only the aops plugin uses `load_pair`. So adding a `verdict.user.md` today would have no effect; those plugins need rewiring first. No test enforces coverage, so a new message can ship without a user line and nothing will flag it.

plugins/cope/README.md
botnicbot: Verified. It describes the system that actually exists. Spot-checked the load-bearing claims against the code: both evaluator protocols, the 4000-character content truncation, the 5-second budget described as a latency ceiling rather than an endpoint, `additionalContext` and never `permissionDecision`, the agy-only `PreInvocation` wiring matching `only_on_clients`, and the API key marked sensitive in the manifest. No regex system is described anywhere in it.

plugins/ida/README.md
botnicbot: Two of the three are real; the hook is not. There is no summarising hook: `plugins/ida/` contains four files — the agent definition, two manifest files, and the README — and the README states "No skills, no commands, no hooks." The don't-parrot behaviour is agent instruction plus included doctrine, not an enforced hook. The earlier reply is honest that the hook approach was abandoned because its output reached the wrong session, but the answer to the question as asked is no. The evidence reminder is real, registered for `Stop` and `SubagentStop`, and lives in the aops core plugin so it applies to every agent rather than only ida. Both bug fixes are real and tested: the undeliverable parent-facing message was deleted rather than left in place, and the `stop_hook_active` guard is present in both shipped artifacts and covered by parametrised tests.

plugins/pkb/agents/pauli.md
botnicbot: Resolved by a ruling from the repository owner: planning uncertain futures is inherently ambiguous, so ordinary planning uncertainty is not a halt. Halt is for when the work does not function, and for when a spike genuinely cannot be run or the alternative branches cannot be modelled. The omitted guidance has been restored to pauli — workers are full-judgment agents, in-repo design ambiguity is not a halt, name the conflict, pick a sensible default, dispatch. `lib/doctrine/halt.md` has been narrowed to match, and the new wording is verified present in all five agents that include it across both client builds. One open item for the owner: `lib/axioms/closure.md` still says nothing may be decided that is not derivable from an axiom, instruction, or user directive, which read strictly covers picking a default for an unspecified implementation detail. A one-sentence reconciliation is proposed but not applied, since axioms are held for explicit sign-off.

plugins/pkb/skills/decompose/SKILL.md
botnicbot: Verified for what the comment anchored on. Step 2 now names all three template layers as one namespace of equal components, with the PKB index as the entry point, and forbids the skill carrying a process in its own text. The workflow skill implements it: three layers, later-wins-by-name, the `wf-*` prefix explicitly a naming convention rather than a privileged set, and drift reported rather than silently repaired. The PKB index resolves live and enumerates 23 templates with zero drift against the tag query. Two things remain hardcoded outside the commented region, flagged rather than changed: Step 3 of the same skill hardcodes the default review set and the human-sign-off line in its own prose, which is the pattern the workflow skill explicitly forbids and which duplicates an existing PKB template; and 28 process templates still ship as files. The shipped library is layer 1 by design and overridable, so it is defensible — but if the intent was to avoid a hardcoded library at all, that is not met.

plugins/pkb/agents/pauli.md
botnicbot: Withdrawing the proposed `closure.md` addition from the comment above. The principle it would have added is already stated, and more forcefully, in `lib/axioms/exercise-authority.md`: "Inside it, refusing to act is abdication" and "Asking permission for a safe, reversible, workflow-required action is the violation, not the safe option." Adding it to `closure.md` would be a second copy of an existing rule, which `single-source-of-truth` forbids. No axiom change is needed. The two axioms already resolve the tension jointly: closure forbids inventing authorisation you were not given, exercise-authority requires you to make the implementation choices inside work you were authorised to do.

nicsuzor (approved):

PR #2345, opened 2026-07-28
v0.6 readmes

botnicbot:
- **so many questions: readme.md notes**
- **more README notes to address**
- **style: autofix lint and formatting**
- **update readme files**


nicsuzor (changes requested):
# Code Review — commit `47069c0` in the `academicOps` repository (main worktree)

These changes introduce a new provenance module and overhaul the documentation for the `aops` and `pkb` plugins. While the code is structurally clean, several issues exist: a potential security/robustness vulnerability with unbounded file reading, arbitrary workflow limitations forbidding agent teams inside containerized workers, and instruction bloat caused by indiscriminate `@include` directives. Additionally, standard diagnostic XML tags were omitted from the injected templates, and user-facing messages should be simplified to a single substantive line.

## **[WARNING]** Insecure handling of potentially large registry files

The _load_json function reads the entire registry file into memory using read_text(). If an attacker or a malformed client environment points this at an unexpectedly large file, the plugin could crash or cause resource exhaustion.

Files:
- `lib/hooks/provenance.py`

### **[WARNING]** Unbounded file read

The use of path.read_text() is unbounded. Consider checking file size before reading or using a streaming JSON parser if the registry file size is not guaranteed to be small.

`lib/hooks/provenance.py:43`

## **[WARNING]** Over-restriction of worker capabilities

The documentation explicitly rules out the use of "teammate mode" or agent teams inside containerized workers. Micromanaging the internal execution strategy of a worker limits the agent's ability to efficiently handle tasks. Once a task is isolated within the container, the agent's choice of internal orchestration should not be artificially constrained.

Files:
- `plugins/aops/README.md`

### **[WARNING]** Artificial restriction on agent teams

Remove the blanket prohibition on agent teams and teammate mode for workers. Workers should be free to utilize available tools and harness features to resolve their tasks effectively once isolated in a container.

`plugins/aops/README.md:68`

## **[WARNING]** Context window bloat from build-time @include directives

Inlining a large list of general doctrine files (such as 'bar', 'launder', 'probe', 'delegation', 'epistemics', etc.) into agent definitions at build time dramatically inflates the prompt size. This increases latency, token costs, and risks diluting the agent's focus on its core instructions. Agent and skill definitions should be highly selective and include only directly applicable doctrine files.

Files:
- `plugins/ida/agents/ida.md`
- `plugins/ida/README.md`

### **[WARNING]** Indiscriminate @include in agent instructions

Including multiple separate doctrine documents like `bar.md` and others listed in the project layout increases instruction size with little benefit. Review the necessity of each included file and consolidate key principles into highly condensed guidelines instead of importing whole documents.

`plugins/ida/agents/ida.md:48`

## **[WARNING]** Restore wrapping markers and optimize user-facing output

Injected hook content must be properly delimited using custom XML-style wrapping tags (e.g., &lt;academicOps session_start&gt;...&lt;/academicOps session_start&gt;) to ensure the client/harness can correctly isolate and parse them. Furthermore, user-facing output should avoid wordy meta-narratives and present actual findings concisely on a single line.

Files:
- `plugins/aops/hooks/messages/session-start.md`
- `plugins/aops/hooks/messages/session-start.user.md`

### **[WARNING]** Missing wrapping markers around injected content

Injected system/agent context in session-start.md should be explicitly enclosed in standard &lt;academicOps session_start&gt;...&lt;/academicOps session_start&gt; block tags so the harness can recognize and process the boundaries of the hook's output.

`plugins/aops/hooks/messages/session-start.md:1-4`

### **[SUGGESTION]** Convert user-facing message to a single-line factual summary

Rewrite the template to output a single, informative line detailing what was actually discovered (e.g., the plugin name, version, and telemetry status) rather than a meta-notification about the agent being messaged.

`plugins/aops/hooks/messages/session-start.user.md:1`

## **[SUGGESTION]** Remove redundant negative constraints

Documentation frequently lists what agents "never" do (e.g., "James never talks to the user"). This creates cognitive overhead; documentation should focus on defining the agents' actual positive responsibilities and contract boundaries rather than enumerating excluded behaviors.

Files:
- `plugins/aops/README.md`
- `plugins/aops/agents/james.md`
- `plugins/ida/README.md`

### **[SUGGESTION]** Negative constraint clutter in Aops README

Remove the phrasing: "James never talks to the user; ida does. He never executes the work either..." and focus solely on his positive role as dispatcher/certifier.

`plugins/aops/README.md:56`

### **[SUGGESTION]** Negative constraint clutter in James agent docs

Remove: "You route work; you never do it." This adds no functional value to an agent's instruction set.

`plugins/aops/agents/james.md:68`

plugins/cope/hooks/evaluator.py
```diff
@@ -138,12 +138,15 @@ def _note(message: str) -> None:
...
-    cloned repository cannot supply one. That makes it the more trustworthy of
-    the two sources, and it wins. The plain environment variable is the fallback
-    that agy and container sessions rely on, neither of which has userConfig at
-    all.
+    Claude Code ignores the ``pluginConfigs`` block — the settings key a
+    ``userConfig`` answer is recorded under — when it appears in a project's
+    ``.claude/settings.json`` or ``.claude/settings.local.json``, a restriction
+    documented as a security measure and specific to ``pluginConfigs`` rather
+    than to settings generally (Claude Code plugins reference, "User
+    configuration"). Both those files live in the workspace, so a cloned
```
botnicbot, line 146: **The causality is inverted.** Living in the workspace is precisely why a cloned repository *can* supply one of these files — that is the hazard this restriction exists to mitigate, not a reason the hazard is absent.

This PR states it correctly 20 lines away, in `plugins/cope/README.md:126`: "both those files live in the workspace, so without it a cloned repository could supply values that flow into plugin hook commands, MCP server configs, LSP commands, and monitor commands."

Adopt the README's wording here.

plugins/aops/README.md
```diff
@@ -2,96 +2,184 @@
...
+
+The review shape does not change with the surface. Cadence is a routing detail; the claim, the return contract, and the certification step below are identical whichever surface ran the work.
+
+### The task is the contract
+
+The boundary is `claim_task` in, `release_task` out — one session, one claimed task, its children included (`specs/enforcement/task-contract.md`). An epic is a task with children and needs no special machinery. The invariant that a session claims exactly one task is held by convention and review; that spec concedes there is no code-level blocking check for it today.
+
+**What the task carries when it is briefed.** The body accumulates, in order: `hydrate`'s four-section context bundle; `decompose`'s record — the subtask DAG, the composed process, and the review obligations emitted as real blocking nodes; then `brief`'s seven elements, appended under what is already there. The brief is composed immediately before dispatch and freshly each time, and the executor is handed a task id rather than the text, so its first act is to read the brief cold. A brief written earlier and unchanged since may be dispatched as it stands.
```
botnicbot, line 82: **Two adjacent sentences negate each other.** "The brief is composed immediately before dispatch and freshly each time" is followed by "A brief written earlier and unchanged since may be dispatched as it stands."

The underlying conflict predates this PR and lives in the skills themselves — `plugins/pkb/skills/brief/SKILL.md:28` ("A brief that is already a stable artifact — written in an earlier invocation, unchanged since — may be dispatched directly") against `plugins/aops/skills/dispatch/SKILL.md:29` ("never earlier, and freshly each time even if that task was briefed on an earlier pass").

Resolve it at the two skills, then state the resolution once here rather than holding both positions in one paragraph.

plugins/aops/README.md
```diff
@@ -2,96 +2,184 @@
...
+
+### What james hands back to ida
+
+Certification is the first of two signatures on `done`; acceptance against the user's intent is ida's, and neither substitutes for the other (`specs/enforcement/workflow.md`). What crosses between them is the evidence contract, defined once in `specs/enforcement/evidence-contract.md`. Its canonical structured form is six fields:
+
 ```
+VERDICT: <PASS | PARTIAL | FAIL | BLOCKED | NEEDS-PRINCIPAL>
+CLAIM: <one sentence — the conclusion>
+GATE: <the acceptance criterion tested, and the observed result against it>
+EVIDENCE: <pointers — command+output, file:line, resolving URL, quoted source>
```
botnicbot, line 116: **A duplicated fact that has already drifted.** Line 110 says the evidence contract is "defined once in `specs/enforcement/evidence-contract.md`" and then reproduces the six-field block anyway. The copy has already lost a constraint: the spec at `specs/enforcement/evidence-contract.md:104` reads `EVIDENCE: <pointers — command+output, file:line, resolving URL, quoted source — NOT pasted dumps>`; this copy drops `— NOT pasted dumps`.

Delete the block and link the spec. That also fixes `plugins/ida/README.md:70`, which currently points readers at this copy.

plugins/aops/README.md
```diff
@@ -2,96 +2,184 @@
...
+    ALLDONE -->|"yes"| BACK["handback to ida:<br/>verdict + evidence + output URL"]
+    VERDICT -->|"REVISE / REJECT"| REWORK
+    VERDICT -->|"APPROVE / MINOR CHANGES"| BACK
     BACK --> IDA
 
-    H1["SessionStart hook<br/>telemetry report + credential scoping"] -.-> PC
-    H2["Stop hook<br/>answer-evidence.md"] -.-> W
-    H3["SubagentStop hook<br/>answer-evidence.md"] -.-> W
-    H4["PreToolUse hook<br/>headless-interactive-prompt.md<br/>refuses a prompt no one can answer"] -.-> W
+    HOOKS["hooks/handlers.py — fires in every Claude Code session,<br/>james's, the team's and the worker's alike.<br/>SessionStart · PreToolUse · Stop · SubagentStop.<br/>Advisory; not a step in this flow."]
```
botnicbot, line 54: **Diagram defects, collected here.**

- This `HOOKS` node is a 25-word label whose final clause is "Advisory; not a step in this flow" — a box inside a flowchart declaring that it is not in the flow. If it is not part of the flow, it should not be a node in it.
- Two separate pauli nodes, `PAULI` at line 11 and `PAULI2` at line 20, with no class, style, or edge marking them as the same agent.
- Line 28 carries a placeholder-shaped node, `OTHER["( another runner )"]`.
- The `IDA` node has exactly two outgoing edges (lines 10 and 11). `plugins/ida/README.md:17` draws a third branch out of ida — `ANS["answer inline"]`, for "a status check, a read, a cheap probe" — which this diagram drops. That is the cross-graph reconciliation one of the deleted review notes asked for.

plugins/cope/README.md
```diff
@@ -100,20 +90,40 @@ Two protocols are supported, so the same mechanism serves a hosted API and a loc
...
+
+### What a roster line contains
+
+One line per live rule, and the line is the rule's frontmatter `description` — nothing else. No rule body reaches the roster, whatever its length: bodies are not truncated or summarised, they are simply not in it. The body has two other destinations, which is why the roster does not need it. It goes to the evaluator as the `policy`, whole; and on a match it goes into the advisory whole, because a rule named without its content is a scolding the agent cannot act on. The roster is composed at hook time from the rule files as they are on disk for that turn, sorted by layer then slug.
+
+The sharp edge is a rule file with `trigger: always_on` and no `description`: its line is a bare slug, a name with no content behind it. Every shipped axiom carries a description, so this can only bite a project- or user-authored rule — which is exactly the kind the roster exists to announce.
+
 ## Configuration
 
 Every value arrives from the environment or from client `userConfig`. No endpoint, host, model name, token, or path is baked into anything this plugin ships, and nothing here has a default.
```
botnicbot, line 115: **"nothing here has a default" is contradicted by this section's own table.** The `COPE_EVALUATOR_TIMEOUT` row at line 134 carries a Default column reading "5.0 seconds", and that default is real — `plugins/cope/hooks/evaluator.py:76`, `DEFAULT_TIMEOUT_SECONDS = 5.0`.

The sentence is defensible about endpoints, hosts, tokens and paths, which is clearly what it means. Scope the claim to those rather than asserting it absolutely two lines above a Default column.

plugins/tools/README.md
```diff
@@ -53,6 +53,23 @@ skill itself — there is no separate router script.
 | `style`         | Builds a personal writing style guide from writing samples.                                         |
 | `new-project`   | Scaffolds a new research project repository end to end.                                             |
 
+## Configuration
+
+No path, host, or credential is written into any file this plugin ships — every
+value below comes from the environment at the point a skill runs. There is one
+fallback: `peer-review` uses `~/brain` when `$ACA_DATA` is unset. This plugin
```
botnicbot, line 60: **A hardcoded path in a shipped artifact.** This paragraph asserts "No path, host, or credential is written into any file this plugin ships" and then names the exception in the next sentence. The exception is real and ships in three places: `${ACA_DATA:-~/brain}` at `plugins/tools/skills/peer-review/SKILL.md:85` and `:195`, and at `plugins/tools/skills/peer-review/references/reading-notes-format.md:20`.

`.agents/CORE.md` forbids any path appearing in a shipped artifact — the fix is to drop the fallback and fail loudly when `$ACA_DATA` is unset, not to document it. Those three files are outside this PR's diff, so this is worth fixing independently of everything else here.

tests/test_aops_handlers.py
```diff
@@ -177,6 +177,74 @@ def test_askuserquestion_is_allowed_when_someone_is_there_to_answer():
...
+
+
+def test_session_start_reports_all_three_facts():
+    """Telemetry was always reported; the build and the installed roster are
+    what the session had no way to see. All three land in one advisory."""
+    res = _session_start()
+    assert res is not None
+    injected = res["inject_text"]
+    assert "telemetry:" in injected
+    assert "plugin:" in injected
```
botnicbot, line 205: **These assertions only pin the degraded path.** `"plugin:" in injected` is equally satisfied by the healthy `f"plugin: {name} {version}"` and by the degraded `"plugin: build manifest not readable"` (`lib/hooks/provenance.py:78`). Same for `"plugins installed:"` against `"plugins installed: no client registry beside this build"` (line 95).

Run against a source checkout, both facts *always* degrade — `_manifest()` needs a rendered `plugin.json` that only `make build` writes, and `_registry()` needs an `installed_plugins.json` that only a real client install writes. Neither exists in the tree. So the mechanism could report "not readable" permanently and these five tests stay green.

`tests/test_shipped_hooks.py` is the suite that builds `dist/` and runs a shipped hook end to end, and it is untouched. That is where the non-degraded values can actually be asserted.

lib/hooks/provenance.py
```diff
@@ -0,0 +1,104 @@
...
+given the location of, and never writes.
+
+Both facts are read from files this runtime is standing in, located by walking
+out from ``__file__``. Nothing here contains a host path, a home directory, or
+a client install path: a shipped artifact may not carry one (specs/
+ARCHITECTURE.md, Binding constraints), and a baked search path is exactly the
+regression ``tests/test_shipped_hooks.py`` already pins for the ts hook. So the
+plugin root is the hooks directory's parent, and the client's plugin registry
+is found by walking up from there until a directory holding it appears. Neither
+lookup is a plugin reading another plugin's files: the first is this plugin's
```
botnicbot, line 12: **The precedent is overclaimed.** `credentials.py` is handed an exact path and does no discovery — `lib/hooks/credentials.py:36`, `env_file = os.environ.get("CLAUDE_ENV_FILE")`, with no walk anywhere in the file. `_registry()` here walks *every* ancestor directory looking for a hardcoded filename (line 67, `for ancestor in Path(hooks_dir).resolve().parents:`, against `_REGISTRY_NAME = "installed_plugins.json"`). Those are not the same class of thing, and the difference is the whole of what the no-defaults rule is about.

Relatedly, `test_provenance_bakes_no_host_or_install_path` (`tests/test_hooks.py:649`) only asserts five home-directory-shaped tokens are absent from the source text. It does not test the registry filename and does not exercise the walk — so it reads as coverage of the no-defaults constraint without being it.

plugins/aops/hooks/messages/session-start.user.md
```diff
@@ -1 +1 @@
-aops: reported this session's telemetry configuration to the agent. Reported only — nothing was set, and no endpoint was supplied.
+aops: reported this session's plugin build, installed plugins, and telemetry configuration to the agent. Reported only — nothing was set, and no endpoint was supplied.
```
botnicbot, line 1: This claims installed plugins were reported, and it says so identically whether `provenance.installed()` returned a real roster or degraded to "no client registry beside this build". Since the degraded case is the *normal* one outside an installed client, the line as written will usually overstate what the session was told.

plugins/aops/hooks/handlers.py
```diff
@@ -60,14 +61,24 @@ def _warn(ctx: HookContext, name: str) -> result.Result:
...
+    scope credentials for container sessions.
 
     Reports only. Never sets a telemetry value and never supplies an endpoint.
-    """
-    # <!-- NS: Make sure this reports plugin version and loaded plugins and whatever else we're going to need -- tracing urls etc. but have a separate hook for functionality that exists only in other packages, don't introduce cross-package messages -->
 
+    Three compact facts, one line each. This fires on every session, so it is
+    held to the injection-tier discipline the axioms impose: a version string
+    and a name list, not a report. A trace URL is not among them and cannot be
+    — see lib/hooks/telemetry.py, ``EXPORT_VARS``.
```
botnicbot, line 72: The conclusion is right — `report()` emits coverage counts, never values — but the pointer is imprecise. `EXPORT_VARS` *does* contain `OTEL_EXPORTER_OTLP_ENDPOINT` (`lib/hooks/telemetry.py:32`). What keeps a trace URL out of the injected line is that `report()` counts against `ENABLEMENT_VARS` only. Point at that instead.

plugins/ida/README.md
```diff
@@ -35,25 +36,46 @@ flowchart TD
...
+
+**Acceptance is ida's**, against the user's intent. Certification without acceptance ships work nobody weighed against what was wanted; acceptance without certification asks her to vouch for mechanics she never saw.
+
+Ida routes but never certifies. The reconcile sweep detects uncertified work but never certifies it either. There is no status and no frontmatter field for certification: it is prose on the task record, and the four surfaces that write it are named in the aops plugin's own README.
+
+What crosses between them is the evidence contract, defined once in `specs/enforcement/evidence-contract.md` and set out with its structured form in `plugins/aops/README.md`, which owns james. Whichever form a handback takes, two rules bind it: checkable evidence or a stated failure reason, and every claim labelled Observed or Reported. It is written to the PKB task record, which is the only message bus; a result held in a transcript or a pull-request body has not been delivered.
```
botnicbot, line 70: This points at `plugins/aops/README.md` as where the evidence contract's structured form is "set out" — but that copy has already drifted from the spec (see the comment on `plugins/aops/README.md:116`). Point at `specs/enforcement/evidence-contract.md` directly.

botnicbot (commented):
## REVISE

**What works.** The prose is genuinely strong, and it is accurate where it stays inside its contract — a broad spot-check of agent/skill bindings, hook names, environment variables and spec paths resolved correctly throughout. The provenance mechanism underneath is not a duplicate path: searching the pre-PR tree for a prior reporter of build identity or the installed roster (`git grep -n -e 'build_manifest' -e 'installed_plugins' 03af217d6 -- 'lib/**' 'plugins/**'`) returns only polecat's container plugin-cache pruning, nothing that reports either fact into a session. The build-time injection is single-copy. The `dispatch/SKILL.md` correction — routing `verify` to the agent that owns it rather than invoking it inline — is a real fix. The module-level provenance tests in `tests/test_hooks.py` are thorough. `make lint` at the head commit exits 0.

The verdict is REVISE for scope and for a set of cross-document contradictions, not for the quality of the writing.

---

### 1. The scope is undisclosed

The title and body say "readmes". The single commit in this PR (`47069c0e8`) also adds `lib/hooks/provenance.py` (new, 104 lines), rewires the `SessionStart` hook to inject two new lines into every session (`plugins/aops/hooks/messages/session-start.md` gains `{plugin}` and `{installed}`), adds 217 lines of tests across two files, and adds a refcheck allowance to `pyproject.toml`. `plugins/aops/skills/dispatch/SKILL.md` separately changes who commissions `verify` — a behavioural change to the review machinery.

The body's four bullets do not describe this. Three of them name commits that are already on `origin/v0.6` and are **not** in this PR (`1a638fcb5`, `5e77aa767`, `03af217d6`); the fourth is the one commit that is here. The body omits the one material behaviour change the PR actually contains.

Recommend splitting: READMEs in one PR, the session-provenance mechanism with its own description in another. `.agents/CORE.md`: "Bundle an unrelated fix only when it blocks your PR from merging and it is one sentence to describe; otherwise file it separately."

### 2. The READMEs are spec content misfiled

`specs/ARCHITECTURE.md:47` sets the plugin-README contract and ends "No history, no rationale, no roadmap, no changelog." The aops, cope, ida and pkb READMEs are rationale end to end. Representative headings: `### Certification, and the gap it lives in` (`plugins/aops/README.md:92`), `### Why the roster is injected rather than baked in` (`plugins/cope/README.md:101`), `### Mutation is a convention, not a boundary` (`plugins/pkb/README.md:74`).

The prose is good. As written it is spec material. See the note at the end of this review before acting on this one.

### 3. This PR deletes all outstanding inline review notes

Five `<!-- NS:` comment blocks exist at both `origin/v0.6` and the merge-base `03af217d6` — one in `plugins/aops/README.md`, two in `plugins/cope/README.md`, one each in `plugins/ida/README.md` and `plugins/pkb/README.md`. The head commit carries zero. They cover, among other things, "review diagram and docs for currency and intent and completeness", "review for completeness and accuracy", "this needs to be reconciled with the other graphs from other plugins", and "this graph is broken". Merging as-is discards unaddressed review feedback rather than answering it.

### 4. Two authoritative specs still describe the pre-PR hook

The hook now reports three facts. Neither spec says so:

- `specs/ARCHITECTURE.md:138` — "Credential isolation for container sessions. Report telemetry configuration." This omits build identity and the roster, and it is separately contradicted by this PR's own `plugins/aops/README.md:170`, which states of that same branch: "**What that branch does is not isolation.** Every value it writes was read from the process environment and stays there".
- `specs/enforcement/enforcement.md:50` — "credential isolation for container sessions, plus a report (never a value) of the OpenTelemetry configuration", introduced at line 48 as "The complete set". It is no longer the complete set. This PR edits that file, but only at line 89 on an unrelated matter.

`.agents/CORE.md`: "A doc that lies is worse than no doc." The project rule that enforcement changes update the spec in the same PR applies directly here.

### 5. Live contradiction on who may write to the PKB — open question, not a wording fix

`specs/ARCHITECTURE.md:98` (untouched): "**pauli** is the sole writer to the PKB. No other agent mutates it."

`plugins/ida/README.md:46` (added here) makes that exclusivity load-bearing for a delegation prohibition: "pauli is the only writer to the store, and a spawn that lands anywhere else reads a graph it cannot correct."

`plugins/pkb/README.md:74-83` (added here) says the opposite: "That is doctrine, not permission ... James holds read, knowledge-write, and task-lifecycle operations outright by its own charter, and routes only graph mutation ... to pauli. Restore the boundary as a tool grant when the harness defect is fixed."

One PR asserts both. This needs a decision about what the boundary actually is before the prose can be made consistent — flagging it rather than proposing an answer.

### 6. The diagrams violate the repo's own mermaid standard, which this PR edits

All five README mermaid blocks contain zero `classDef`, `style`, `linkStyle`, or `%%{init}` directives. `plugins/tools/skills/diagram/references/mermaid-techniques.md` — modified by this PR — mandates a theme init block at line 68 ("## Start with Init Block for Theme and Spacing", "**CRITICAL**: Proper spacing is the #1 factor for readability") and at line 54 requires "Summarize verbose labels to 3-5 words; let shape convey additional context." The longest label in `plugins/aops/README.md` is 25 words. Worth noting that this PR's edit to that reference file changed its `classDef`/`%%{init}` example fences from ```` ```mermaid ```` to ```` ```text ````.

Diagram-specific defects are commented inline.

### 7. A dead path inside a mermaid diagram, and the guard that cannot see it

`plugins/tools/README.md:14` draws `TD["extract/workflows/training-data.md"]`. That file does not exist: `plugins/tools/skills/extract/workflows/` contains only `docs-to-md.md`, and the training-data workflow lives inline at `plugins/tools/skills/extract/SKILL.md:98`.

The systemic part is worth its own issue. `scripts/check_refs.py` blanks fenced blocks before extracting path candidates (`mask_fences`, lines 78-92, on the rationale that "Fenced blocks are illustrative by construction"). Every mandated README mermaid diagram is therefore exempt from path checking — `make lint` reports "every documented repository path exists" and exits 0 while this broken path ships.

### 8. The branch is red and needs a rebase

The head commit is 5 commits behind `origin/v0.6`. Collecting the suite at the head commit exits 2, on `tests/polecat/test_live_edit_mount.py` — `AttributeError: module 'aops.polecat.cli' has no attribute 'CONTAINER_PLUGIN_CACHE_ROOT'`. That file no longer exists on `origin/v0.6`; the PR touches neither it nor `cli.py`, so this is inherited from the merge-base and resolved by rebasing, not a defect introduced here. (A second collection error, `tests/test_plugin_manifests.py` requiring `make build`, is an artifact of the unbuilt checkout I collected in, not of the branch.)

---

### Open questions for you, not change requests

- The PKB-writer boundary above (#5).
- Whether the installed-plugin roster belongs in every session's context at all, or in the structured hook log instead.

### One note on #2 and #6

You have ruled on what a plugin README is: it must show the user exactly how the plugin works, and does not need technical source citations. `specs/meta/doc-taxonomy.md` does not yet carry that definition — it is being updated separately, and today it mentions only top-level `README.md` at line 67. Findings #2 and #6 should be actioned against that standard once it lands, not against the current `ARCHITECTURE.md` wording.

PR #2403, opened 2026-08-07
fix(transcripts): stop HTML escaping from defeating credential redaction (#2373 remediation)

botnicbot:
> **Before you do anything else — the order matters.**
>
> **Merge this PR first, then `git pull` in `/home/nic/src/academicOps`.** That checkout is pinned behind merge `ad7e544f5` and still runs the safe pre-merge renderer. Pulling before this lands is what turns a latent credential leak into a live one, within one cron cycle. **Nothing has leaked.**
>
> **Also: do not run the transcript renderer from any other worktree.** Three non-host worktrees carry the merge on disk; none of them renders transcripts today.
>
> Full detail under **Exposure window** at the end of this body.

Remediates the blocking findings from the review of #2373, which was merged into `v7.1` at `ad7e544f5` while its review was still running. Repair-in-place on a fresh branch off `v7.1`, per that review's remediation plan.

## 1. Credential leak (the reason this is urgent)

`_escape_html` was widened to `html.escape(str(text), quote=True)`. Redaction runs *after* rendering, at the write chokepoint in `runner.py:200-206`, so by the time `redact_secrets` sees the text every `"` is already `&quot;` — which breaks the `\s*:\s*` match in `secret_redaction.py:93`. A key-named secret whose value carries no recognisable token shape (no `ghp_`, no `sk-ant-`) therefore survived into shipped artifacts.

**Precisely which content, because the obvious phrasing overstates it.** `_format_tool_output_markdown` (`renderer.py:242-263`) and the tool-call-args block call **no escaper on either tree** — so in the Markdown tiers, tool output was never affected, and the classic env-dump path that `secret_redaction.py:6-8` records as the 2026-06-01 incident kept redacting correctly throughout. In HTML, tool output *was* escaped (`renderer.py:641`, `:671` at base) and so was hit.

Net: the regression bit **assistant text, human prompts, thinking blocks and table cells across all four text tiers, and tool output only in `.html`** — which was the worst-affected tier. That does not lower the urgency; it is simply what the code does.

Fixed by splitting the tiers rather than by touching the redaction regex:

- **Markdown tiers carry text verbatim** (`_md_text`). Markdown is not HTML. Escaping there also turned fenced code blocks into `&lt;`-noise, which is the second finding below.
- **The HTML tier keeps `quote=True`** — dropping it would re-open attribute injection — but **redacts before escaping**. The chokepoint pass remains, now as defence in depth rather than the only line.

**Proven, not asserted — and proven in the order that matters.** The new test drives the real renderer through `process_single_session` with a deliberately shapeless value. It was run against the *pre-fix* code first and observed to fail, then against the fix:

```
# against the merged (pre-fix) renderer — verified RED first
E   AssertionError: key-named secret leaked into 20260701-10-adhoc-11111111.html
1 failed

# against this branch
33 passed                                     (exit 0)
```

A test written after a fix and never seen red proves nothing about what it detects. This one was.

**The test is the most valuable thing in this branch, and the reason the defect shipped is that a test already existed and could not see it.** `test_runner_scrubs_every_session_artifact` asserts on bytes written to disk — exactly the right shape — but it does this first:

```python
monkeypatch.setattr(runner, "render_session_to_all_formats", lambda *a, **k: (poisoned, ...))
```

It stubs out the very function whose behaviour broke. The renderer never runs, so the escaping that defeats redaction never happens, and the test passes on code that leaks. A chokepoint test that mocks away everything upstream of the chokepoint proves only that the chokepoint calls a function.

The new test removes the stub: real renderer, real `process_single_session`, real files, and a **shapeless** secret value (no `ghp_`/`sk-ant-` prefix) so the token-shape patterns cannot mask a regression and only the name-based rule can catch it. It was verified to fail on the merged renderer *before* it was made to pass — the order matters, because a test written after the fix and never seen red proves nothing about what it detects.

Whatever happens to the rest of this branch, that test should survive.

## 2. Markdown corruption, and the tests that pinned it

Four assertions across three files required `&lt;tag&gt;` *in a `.md` tier*. They encoded the corruption as the contract. They now assert the author's bytes.

## 3. The XSS the test suite required to stay open

`test_r4_adversarial_stress.py:269-270` asserted `has_source_xss is True or has_ts_xss is True` — the suite demanded a live hole. `event.source` and `event.timestamp` were interpolated into the HTML event header unescaped, one of them inside a `class` attribute.

Closing it was natural to this change (the escaping layer was already open), so it is closed here rather than parked: both fields are escaped at emission, and the test is inverted and renamed to guard the fix.

## 4. Cross-plugin read in `lib/hooks/dispatch.py`

On `ImportError` it walked to `plugins/rbg/hooks` and imported `evaluator_otel_trace` from there — a plugin reading another plugin's files, and a path that resolves to nothing in every built artifact, making the instrumentation a silent no-op everywhere but rbg.

The traversal is gone. The sibling import stays, which is the legitimate in-plugin path and is how rbg gets the module. Silence is correct when nobody asked for OTel; when `COPE_EVALUATOR_OTEL_TRACE_PATH` is set and the module is absent, it now says so on stderr instead of vanishing.

## 5. Specs made untrue by the merge

- `specs/transcript-pipeline.md` still described **three** output artifacts; the pipeline writes **five**. The three Markdown tiers are now named by what each carries, and the escaping/redaction contract is recorded — a constraint a future renderer must honour, not an implementation detail.
- `specs/polecat/tmux-interactive-driving.md` promised the host log directory is named by whatever you pass to `-s`. `cli.py` now sanitizes that value, so the promise holds only for names already inside `[a-zA-Z0-9_.-]`. Said where the claim is made.

## The chokepoint is not bypassed — established from produced artifacts

The obvious hazard in this fix: redaction now also runs *inside* `_escape_html`, while the Markdown call sites were repointed *off* that helper onto `_md_text`. If the post-render pass at `runner.py:200-206` had been removed or bypassed, the Markdown tiers would be redacted by nothing — the original bug's exact shape, in a new place.

It has not been. **This branch does not touch `runner.py` at all** (`git diff --stat origin/v7.1...HEAD -- lib/py/transcripts/runner.py` → empty), and the behaviour is confirmed from written bytes rather than from reading code:

```
[0] _md_text is a passthrough (redacts nothing): True
    _md_text output still contains the raw secret: True
[3] ...controller.md   carries_content=True  raw_secret=False  REDACTED=True
    ...full.md         carries_content=True  raw_secret=False  REDACTED=True
    ...md              carries_content=True  raw_secret=False  REDACTED=True
    ...html            carries_content=True  raw_secret=False  REDACTED=True
    ...json            carries_content=True  raw_secret=False  REDACTED=True
[4] artifacts leaking the raw secret: NONE
[5] ...controller.md verbatim line:
    {"MY_API_KEY": "[REDACTED]", "note": "config"}
```

`_md_text` provably removes nothing, yet the secret is gone from every Markdown tier — so the post-render pass is demonstrably still running and is what scrubs them. It can now *see* the `"KEY": "value"` shape precisely because Markdown is carried verbatim. That is the fix working as designed, not around it.

**Counted against the pre-merge baseline**, which was known-good:

| Tier | Pre-merge | Merged (leaking) | This branch |
|---|---|---|---|
| Markdown | 1 pass (chokepoint); escaped `& < >`, not `"` | 1 pass, **defeated** by `&quot;` | 1 pass, on pristine text |
| HTML | 1 pass (chokepoint) | 1 pass, **defeated** by `&quot;` | 2 passes (redact-then-escape, plus chokepoint) |

Pre-merge `_escape_html` was `text.replace("&", "&amp;").replace("<", "&lt;").replace(">", "&gt;")` — it never redacted, and never escaped `"`, which is exactly why the `"KEY": "value"` match held. So the Markdown tiers here are on the same single pass they always had, now seeing *unescaped* text rather than partially-escaped text — strictly better than the baseline, not weaker. The HTML tier gains a pass it never had.

## Test evidence

`tests/transcripts/` — the suite covering everything this branch changes — is **deterministic**:

```
149 passed  (five consecutive runs, -p no:randomly)
```

`tests/test_cope.py` → `125 passed` (exit 0). Reported separately because the transcripts suite does **not** cover `dispatch.py`, so "the tests pass" needs both.

`uv run ruff check lib/ tests/` → `All checks passed!` (exit 0).
`uv run basedpyright lib/py/transcripts/domain/renderer.py lib/hooks/dispatch.py` → `0 errors, 0 warnings, 0 notes` (exit 0). **Run directly**, not via `make lint` — `make lint` aborts at the reference check before basedpyright is reached, so no basedpyright result can be attributed to that command.

**On the full suite, stated with its nondeterminism named.** `make test` does not return a stable failure count at a fixed SHA — this branch produced 18 and then 17 failures across two runs of identical code, and the same instability is reported at other SHAs. A `comm -13` against a single base run was empty both times, but **that comparison is not reproducible evidence** and this PR does not rest on it. What is reproducible: the deterministic suite above, and the leak test's fail-on-merged / pass-on-fixed result. The instability is concentrated in `tests/test_shipped_hooks.py`, which yields 0 failures-with-errors in isolation across three consecutive runs; it is pre-existing and deserves its own issue.

Two base failures do not recur on head (`test_rbg_shipped_hook_traces_every_rule_evaluated_not_only_the_matches`, `test_ts_session_end_no_op_when_not_a_remote_session`). **This branch claims no credit for those** — they live in that same unstable file and flapped; they were not repaired.

`make lint` exits non-zero on **67 pre-existing broken references, all under `specs/future/`** — a directory this branch does not touch, and this branch adds no references. Ruff, the first stage, passes.

## Why Markdown-verbatim is right, stated durably

The strongest justification is not the redaction bug — it is corpus auditability, and it holds even if the redactor is fixed tomorrow.

The recorded design (`aops-00c0fa10`, Nic's framing) puts three layers over secrets — a pre-tool guard, this write-time scrub, and a pre-commit backstop — all sharing **one** pattern definition. That presupposes the committed corpus stays scannable by those patterns *after the fact*. `_SENSITIVE_NAME` is a character class admitting no `&` or `;`, so entity-escaped text is structurally invisible to it and to **any future scanner sharing the definition**. Escaping a tier does not merely inconvenience today's redactor; it makes that tier permanently un-rescannable.

Two supporting points:

- **Escaping Markdown was never a security control.** `html.escape(..., quote=True)` leaves `[click](javascript:alert(1))` byte-identical. It blocked one injection vector while leaving a co-located one open — the appearance of safety, which is worse than none, because it licenses "the `.md` is escaped, so it's safe to render."
- **No exposure class was created here.** The bulkiest, most hostile-influenced Markdown surface — tool output (`renderer.py:242-263`) and tool-call args — was interpolated raw between fences on `origin/v7.1` too, calling no escaper. This PR removed an inconsistency, not a defence.

## Adversarial QA: measured, and the absolute claim withdrawn

A 29-case adversarial corpus was driven through the real `process_single_session`, scanning all five written artifacts for a shapeless secret (`Zq7wPlainShapelessValue42` — no `ghp_`/`sk-ant-` prefix, so only the name-based rule can catch it). Leak cells, case × artifact:

| Tree | Leaking cells |
|---|---|
| base `origin/v7.1` (`3e20a5f`) | **78** |
| HEAD `fix/2373-credential-leak` | **28** |
| HEAD, mutated: chokepoint redaction stripped from the three Markdown writes only | **82** |

Three things follow, and the third is the important one:

1. **No cell leaks at HEAD that is clean at base.** Strictly monotone improvement; no new leak shape introduced.
2. **The write chokepoint is present, reached, and load-bearing.** The mutant — chokepoint removed from `runner.py:200-204` only, in a scratch copy, tracked tree untouched — goes 28 → 82. Nothing else scrubs Markdown; `_md_text` is a verified passthrough. This is the direct answer to "did moving the Markdown call sites off `_escape_html` leave them unredacted": it did not, and removing the chokepoint proves what would happen if it had.
3. **The absolute claim is withdrawn.** An earlier draft of this PR said a key-named secret "can no longer survive into any artifact." **That is false — 28 cells still leak.** The honest statement is 78 → 28, with the entire residue pre-existing: every one of the 28 also leaks on base. They are `redact_secrets` pattern gaps, not this change's doing, and they are filed separately (see below).

Also verified: XSS closure is complete across 12 hostile fields — zero unescaped in the DOM, zero `<script`, zero `" on<handler>=`. No Markdown collateral damage — pipes still escaped, fences still widened, front matter intact, all 29 sidecars parse. And the fence-corruption claim is empirically true: base renders `if a &lt; b and c &gt; d:` *inside a ```python fence*; HEAD renders the source.

## The chokepoint is not redundant

Worth stating so nobody later reads it as dead code and removes it. Its purpose is the *future* renderer that interpolates raw text and forgets to escape — that text arrives at the chokepoint untransformed and matchable, which is exactly the case the mutation test above shows it catching. "Defence in depth" is accurate rather than decorative.

## Dispositions — recorded, not silently dropped

- **`97138d6cf` (the Claude-side `rglob`): DEFERRED, not done.** The #2373 synthesis named two alignments as satisfying the premise objections; the `kb_024d6843` OTel one is delivered, this one is not. `runner.py:47` still reads `for p in claude_dir.rglob("*.jsonl")`. Deferred because the measured gain was zero (774 files discovered either way), so it is a correctness-of-shape change with no behavioural difference, and this branch is scoped to the credential leak. It is **not** abandoned — it needs its own change, and this line records that it was decided rather than forgotten.
- **`controller_*` / `subagent_*` front-matter keys are undefined in the spec**, and `model.py:115-128` makes `controller_tokens`/`controller_cost_usd` pure aliases of `tokens_used`/`cost_usd` — the artifact prints one number twice under two names. Noted, not chased here; documenting them properly would force the aliasing question, which is a design call for Nic.
- **What this fix does NOT cover: [#2405](https://github.com/nicsuzor/academicOps/issues/2405).** The 28 residual leak cells are five `redact_secrets` pattern gaps — a value containing a space (the most realistic), entity-encoded source content, escaped quotes, a fullwidth colon, and a bare value repeated before its assignment. All pre-existing, all live today. **Read that issue before trusting this fix to cover a case it does not.**
- **[#2406](https://github.com/nicsuzor/academicOps/issues/2406) — HTML-tier redaction destroys token metrics.** `input_tokens: [REDACTED]` on benign input, because `\n` has become `<br>` before the chokepoint reads it, so the numeric-sparing guard's `fullmatch` fails. Identical on base. Same root shape as this PR's bug — judgment made on text a renderer already transformed — in the opposite direction.
- **Layer 2 of the recorded secret design does not exist in this repository.** `check_transcript_secrets.py` and `install_sessions_hooks.sh` are absent from this tree, while `secret_redaction.py:13-14` still asserts a pre-commit linter as the complementary backstop. That makes this renderer the **only** line today, which is why the auditability argument above carries weight. Pre-existing, not introduced here, and bounded: the sessions repo itself was not inspected, so an installed hook could survive on the host.
- **`dispatch.py` warning frequency — the one item on this list that is new in this PR.** The stderr warning fires from a function called once per hook event, so with `COPE_EVALUATOR_OTEL_TRACE_PATH` set it would emit on every hook event across the four plugins that do not ship the module. Left as-is under an explicit stop-work instruction, and flagged here rather than buried: once-per-process would be the fix, and it is a two-line change whenever someone picks it up.
- **`renderer.py:134`** passes `_md_text(started)` into a table cell without the `.replace("|", "\\|")` its sibling applies. Pre-existing — `_escape_html` did not escape `|` either — so not a regression.
- **Framing, for a future reader:** redact-before-transform is written above as the general rule, but the enforcement currently lives inside a helper named `_escape_html`, whose name promises escaping and does not advertise that it performs a security scrub. No bare `html.escape` survives in `lib/py/transcripts/` outside that line, so the funnel holds today.

## Explicitly not here

- **The `plugins/rbg/hooks/handlers.py` re-enable** is *not* remediated and is *not* reverted here. It re-armed all four rbg handlers at once and carries its own decision brief for Nic. The ~30 tests that depend on it are untouched.
- **Adapter test coverage** and the **FIX tail** from the posted review are follow-ups, deliberately left out to keep this diff to the blocking findings.
- **The `NormalizedSession` root-cause fix** — redacting the model before rendering, keeping the chokepoint as backstop — would close this class rather than this instance. Reviewed and endorsed as the right deferral for an overnight security fix, not a shortcut. It is the natural follow-up to [#2405](https://github.com/nicsuzor/academicOps/issues/2405).
- The PR body claim on #2373 of "394 passed" still does not reproduce; correcting a merged PR's body is not attempted here.

## Exposure window

Latent, never live, and **nothing has leaked**. The host that renders transcripts (`/home/nic/src/academicOps`) sits on `v7.1` at `6374859f2`, which does not contain the merge (`git merge-base --is-ancestor ad7e544f5 6374859f2` → rc=1), and still carries the safe pre-merge `_escape_html` on disk. No `.controller.md` — a tier the merged renderer writes unconditionally — exists anywhere in the sessions corpus.

Nothing pulls that checkout automatically: the 5-minute cron runs `git fetch origin --prune` only (no `pull`/`checkout`/`reset`/`merge`/`rebase` anywhere in `scripts/repo-sync-cron.sh`), there are no user systemd timers, and no scheduled workflow renders transcripts.

**Therefore the ordering matters: merge this PR first, then pull the render host.** Pulling first is what would turn the latent bug into a live leak, within one cron cycle.

Draft, and holding: nothing merges tonight.


botnicbot (commented):
## Strategic review — REVISE (findings addressed in `7dc48fe44`)

> **On the review state:** filed as a comment-review, not an approval. GitHub refuses `REQUEST_CHANGES` on your own PR (HTTP 422), and every agent here authenticates as `botnicbot`, this PR's author. The intended state is **revise**; nothing below should be read as approval. Same fallback as the #2373 review.

Three reviewers read this branch independently and blind to each other at head `2b8c8c490`. Both returned **REVISE**; neither found anything blocking the security fix. The FIX-grade findings are now addressed in `7dc48fe44`; the rest are recorded in the PR body with explicit dispositions.

### The premise was tested, and it holds

The one question worth asking of this fix was whether "Markdown carries text verbatim" trades a redaction bug for a stored-XSS surface. **It does not**, on three grounds — the third of which was missing from this PR and is now its central justification:

1. The bulkiest attacker-influenced Markdown surface — tool output at `renderer.py:242-263` and tool-call args — was **already** interpolated raw between fences on `origin/v7.1`, calling no escaper. This PR removed an inconsistency, not a defence.
2. What was removed was never a security control: `html.escape(..., quote=True)` leaves `[click](javascript:alert(1))` byte-identical. It blocked one vector while leaving a co-located one open — the appearance of safety, which licenses "the `.md` is escaped, so it's safe to render."
3. **An entity-escaped corpus is permanently unauditable.** `aops-00c0fa10` records a three-layer design sharing *one* pattern definition, which presupposes the committed corpus stays scannable after the fact. `_SENSITIVE_NAME` admits no `&` or `;`, so escaped text is structurally invisible to it and to any future scanner. This holds even if the redactor is fixed tomorrow, which is why it now owns the spec.

### The defect is reproduced, not argued

Independently, against the real module:

```
RAW  -> {"MY_API_KEY": "[REDACTED]"}
ESC  -> {&quot;MY_API_KEY&quot;: &quot;plainvalue123notatokenshape&quot;}
LEAK_AFTER_ESCAPE = True
```

And the chokepoint is confirmed **not** bypassed, from produced artifacts: `_md_text` is a verified passthrough that removes nothing, yet all three Markdown tiers land as `{"MY_API_KEY": "[REDACTED]", ...}`. `runner.py` is untouched by this branch.

### Findings and dispositions

| Reviewer | Issue | Severity | Disposition |
|---|---|---|---|
| rbg, pauli | Escaping principle stated in full in 4 files; already drifted ("producing" vs "building" a DOM) | FIX | **Fixed** — spec owns it, 3 sites carry pointers |
| pauli, team-lead | Rationale canonised a defect (the `&quot;` blindness) as a rule | FIX | **Fixed** — rests on Markdown≠HTML + auditability; defect moved to a dated note at the mitigation |
| rbg | `runner.py:194-199` asserts "a new renderer cannot accidentally ship an unredacted format" — disproved by this PR | FIX | **Fixed**, with the disproof dated |
| pauli, team-lead | `assert len(artifacts) == 5` would turn a *security* test red when `aops_22c422dc` lands | FIX | **Fixed** — asserts the five tiers exist, not that nothing else does |
| rbg | Rule written as an HTML carve-out; generalises to any re-encoding | REVISE | **Fixed** — stated as the general rule; URL-encoding, base64, a future md→HTML converter all break it identically |
| rbg | `dispatch.py` docstring over-claimed "nowhere else" / "every built artifact" | ADVISORY | **Fixed** — narrowed to what was checked |
| pauli | `97138d6cf` glob had no disposition | REVISE | **Recorded as deferred, with reason** — measured gain was zero (774 vs 774) |
| pauli | `controller_*`/`subagent_*` undefined in spec; pure aliases | REVISE | Recorded; documenting them forces an aliasing design call |
| pauli | Layer 2 (pre-commit backstop) absent from this repo while `secret_redaction.py:13-14` claims it | ADVISORY | Recorded — pre-existing, own issue; bounded to this repo |
| rbg | stderr warning fires once per hook event across 4 plugins | ADVISORY | Recorded, **flagged as the one new-in-this-PR item**; once-per-process is the fix |
| rbg | `_escape_html` performs a security scrub without its name saying so | ADVISORY | Recorded; no bare `html.escape` survives outside it, so the funnel holds today |

### Cleared explicitly

- **`_md_text` beside `_escape_html` is a legitimate split**, not duplication — zero shared logic, disjoint call sites, two contracts.
- **The `dispatch.py` fix is rule-compliant.** The entry point puts the module's own directory on `sys.path`, so the sibling import reads *its own plugin's* directory, whichever that is. The removed traversal named another plugin by path — that was the violation.
- **"A defect is never a rule" is not violated** — the Markdown-verbatim rule stands on fence corruption alone and survives a hardened redactor.
- **Both spec claims verified true of the tree**, and neither spec edit carries changelog, history, or deprecation shapes.

### Verification

`tests/transcripts/` → `149 passed` (deterministic, five consecutive runs). `tests/test_cope.py` → `125 passed` — reported separately because the transcripts suite does not cover `dispatch.py`. `ruff check lib/ tests/` → clean. `basedpyright` on all three changed modules → `0 errors, 0 warnings, 0 notes`, run directly (`make lint` aborts at the reference check before basedpyright is reached).

**`make test` is nondeterministic at a fixed SHA** — 18 vs 17 failures across two runs of identical code. No claim in this PR rests on a full-suite failure delta; that evidence was withdrawn rather than defended.

### Verdict

**REVISE, and the revisions are in `7dc48fe44`.** The security substance was sound at first review and is unchanged; every FIX-grade finding was prose, placement, or a test assertion that would have misfired later. Staying draft — nothing merges tonight, and the merge-before-pull ordering in the PR body is the thing to honour in the morning.

nicsuzor (approved):

PR #2425, opened 2026-08-11
feat(orchestrate): import Claude Code OTel gRPC tracer into orchestrate plugin

botnicbot:
## Summary
- Imports `claude_code_tracer.py` into `plugins/orchestrate` with gRPC OTLP binary export support (`insecure=True` for plain-text gRPC endpoints).
- Registers `UserPromptSubmit`, `PreToolUse`, `PostToolUse`, `PostToolUseFailure`, and `Stop` hook handlers in `plugins/orchestrate`.
- Declares `opentelemetry-exporter-otlp-proto-grpc` in `plugins/orchestrate/pyproject.toml` so dependencies ship with plugin distributions.
- Adds unit test suite and live collector integration test (`test_live_collector.py`) reading `OTEL_EXPORTER_OTLP_ENDPOINT` dynamically.
- Refines James orchestrator prompt rules to delegate QA testing to Marsha and rule compliance checking to RBG.

## Verification
- `uv run pytest plugins/orchestrate/tests/`: 16 passed.
- `make build install-dev`: Built and deployed cleanly to system plugins.


build/marketplace.toml
```diff
@@ -44,9 +44,3 @@ directory = "tools"
 name = "tools"
 description = "Domain research skills."
 category = "productivity"
-
-[[plugins]]
-directory = "aops-debug"
-name = "aops-debug"
```
nicsuzor, line 50: don't delete this plugin!

plugins/orchestrate/hooks/handlers.py
```diff
@@ -2,12 +2,28 @@
...
 from collections.abc import Callable
+from pathlib import Path
+from typing import Any
 
 from dispatch import HookContext, Result, load_message_pair, warn
 
+# Ensure parent directory of hooks (plugins/orchestrate) is in sys.path so claude_code_tracer can be imported
+_PLUGIN_ROOT = Path(__file__).resolve().parent.parent
+if str(_PLUGIN_ROOT) not in sys.path:
+    sys.path.insert(0, str(_PLUGIN_ROOT))
```
nicsuzor: i don't think this is needed -- or it shouldn't be, because the common dispatch.py code should run from the correct dir. if our pyproject.toml is not in a correct dir, fix that rather than work around it.

plugins/orchestrate/manifest/hooks.template.json
```diff
@@ -50,6 +93,16 @@
     },
     "agy": {
       "hooks": {
+        "PreInvocation": [
+          {
+            "hooks": [
+              {
+                "type": "command",
+                "command": "uv run --project \"${AGY_PLUGIN_ROOT}\" python3 \"${AGY_PLUGIN_ROOT}/hooks/dispatch.py\" agy PreInvocation"
```
nicsuzor, line 101: add the other equivalent agy hooks here too: PreToolUse and PostToolUse, PreInvocation, PostInvocation, and Stop.

plugins/orchestrate/hooks/handlers.py
```diff
@@ -86,9 +102,99 @@ def honest_output(ctx: HookContext) -> Result | None:
...
+    except Exception as exc:
+        log.warning("stop tracer failed: %s", exc)
+    return None
+
+
 HANDLERS: dict[str, list] = {
     "SessionStart": [session_start],
+    "UserPromptSubmit": [user_prompt_submit],
+    "PreToolUse": [pre_tool],
+    "PostToolUse": [post_tool],
```
nicsuzor, line 188: verify these get translated to agy equivalents

plugins/orchestrate/hooks/handlers.py
nicsuzor: or move claude_code_tracer.py into the common libs dir if needed. particularly if that helps us avoid having multiple pyproject.toml files

nicsuzor (changes requested):

